PR #30, opened 2018-09-10
Sort atoms by plane, then layer

DamianX:
Not sure if there's a more idiomatic way to code this


SpaceManiac:
`sort_by_key(|a| (plane_of(objtree, a), layer_of(objtree, a)))` would be cleaner than calling `cmp` manually.

I think it's also worth noting that planes are defined by BYOND to be integers from -100 to 100, so handling floats might not be necessary.

SpaceManiac:
It's a cheap hack to convert `f32` into an `i32` which can be sorted on (`f32` is not `Ord` and thus not sortable because `NaN` exists).

PR #40, opened 2018-10-10
Make FILEID_ constants public.

PJB3005:


SpaceManiac:
Not against this but do you have a usecase for `_MIN` or `_MAX`? The other constants are already accessible.

PJB3005:
I.. did not realize there's a `builtins` function. Whoops.

PR #41, opened 2018-10-10
ProcValue now contains the proc AST.

PJB3005:
I need this for retrieving the AST for compiling it.


SpaceManiac:
This does need to be exposed but I think this is the wrong place to do it. I would expect that people want to look up proc bodies by their name (and thus find it on the `ProcValue` struct) rather than by the location in the code where they appear.

PJB3005:
@SpaceManiac fixed.

PR #42, opened 2018-10-10
Actually emit Term::Null and Term::ReturnValue.

PJB3005:
They were emitted as Term::Ident("null") and Term::Ident(".")

### DIFF
--- a/src/dreammaker/parser.rs
+++ b/src/dreammaker/parser.rs
@@ -1594,6 +1594,10 @@ impl<'ctx, 'an, I> Parser<'ctx, 'an, I> where
                 None => Term::Ident(i.to_owned())
             },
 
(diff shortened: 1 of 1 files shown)


SpaceManiac:
Decided to just remove `Term::ReturnValue` in favor of `Ident(".")`. Makes it slightly easier to deal with  stuff like this down the line:

```dm
/obj/item/reagent_containers/food/snacks/customizable/Destroy()
	for(. in ingredients)
		qdel(.)
	return ..()
```

Going to let `Term::Null` through and officially file this (working) DM code in the "too stupid to parse" category:

```dm
/client/proc/foo(null)
    null += 200
    src << null
/client/New()
    ..()
    foo(50)  // 250
```

PR #71, opened 2019-03-05
Add database procs to builtins

DamianX:


SpaceManiac:
Do you have code that calls these directly?

DamianX:
Yeah, this is Dantom.DB https://github.com/vgstation-coders/vgstation13/blob/4cc066f3b557ca5ff239511d628e33520ccfc3c9/code/libs/db/core.dm#L32

PR #89, opened 2019-07-07
Add built in values for most of the atom vars

SpaiR:
It seems legit, that vars with default values will have those values. I also use them in my editor, so it's time to move them from my local copy to original repo.

Values themself are taken from official BYOND ref.


SpaiR:
@SpaceManiac I see. I actually didn't know about that behaviour, so yeah. I removed these definitions. Please check last commit.

PR #102, opened 2019-07-25
Very basic dark theme, use better fonts.

PJB3005:
Dark theme uses prefers-color-scheme media query.
Fonts are Roboto and Source Code Pro from Google Fonts.


SpaceManiac:
Thanks for the dark theme, still deciding if I want to pull in fonts or not though

SpaceManiac:
Merged as 89eafa9e14091f65b42ce3cacce2e2e95fa553ab

PR #108, opened 2019-11-04
Drop proc AST earlier to avoid memory fragmentation.

PJB3005:
This gave me about a ~200 MB saving on Linux when running the lang server against /vg/.


PJB3005:
Welp, apparently I messed up my testing and this doesn't actually work, damnit.

PR #114, opened 2019-11-23
must_not_override

spookydonut:
This is meant to add a linter hint to prevent procs being overridden.

i hope to also add something similar for variables

This is basically `final` in other languages

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -326,6 +326,7 @@ impl<'o> AnalyzeObjectTree<'o> {
             objtree,
             return_type,
             must_call_parent: Default::default(),
+            must_not_override: Default::default(),
             used_kwargs: Default::default(),
         }
     }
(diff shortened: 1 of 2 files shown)


spookydonut:
Tested this and it works

src/dreamchecker/lib.rs
```diff
@@ -542,16 +557,24 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
...
+        if self.proc_ref.parent_proc().is_some() {
             //eprintln!("{:?}", self.env.must_call_parent);
-            let mut next = Some(self.proc_ref);
+            let mut next = Some(self.proc_ref); 
             while let Some(current) = next {
-                if let Some(&must) = self.env.must_call_parent.get(&current) {
-                    if must {
-                        error(self.proc_ref.location, format!("proc never calls parent, required by {}", current))
+                if let Some(&musnt) = self.env.must_not_override.get(&current) {
+                    if musnt {
```
SpaceManiac: I'm not sure exactly mirroring the structure for `should_call_parent` is appropriate. In the case of `should_call_parent`, we want to be able to suppress the warning if a child proc has a good reason not to call parent. Having a 'final' proc that isn't actually final because you can override it by saying you really meant to doesn't seem as useful to me.

spookydonut: @SpaceManiac what if I made it so you can't set it false but kept this code?

spookydonut: You can no longer do SHOULD_NOT_OVERRIDE(FALSE)

PR #115, opened 2019-11-26
Final vars

spookydonut:
adds final vars, `var/final/foo` similar to how `var/static/foo` etc work.

This also checks for duplicate definitions of the same var eg;
```
/a
  var/foo
/a/b
  var/foo
```
will now raise an error in dreamchecker.

### DIFF
--- a/src/dreamchecker/README.md
+++ b/src/dreamchecker/README.md
@@ -54,11 +54,13 @@ be enabled:
 	#define SHOULD_CALL_PARENT(X) set SpacemanDMM_should_call_parent = X
 	#define UNLINT(X) SpacemanDMM_unlint(X)
 	#define SHOULD_NOT_OVERRIDE(X) set SpacemanDMM_should_not_override = X
+  #define FINAL_VAR var/final
 #else
 	#define RETURN_TYPE(X)
 	#define SHOULD_CALL_PARENT(X)
 	#define UNLINT(X) X
 	#define SHOULD_NOT_OVERRIDE(X)
+  #define FINAL_VAR var
 #endif
 ```
 
@@ -85,6 +87,15 @@ parent proc is being skipped. Child procs may set this setting to `0` instead
 to override the check.
 
 ### Should not override
+
 Use `set SpacemanDMM_should_not_override = 1` to raise a warning for any child
 procs that override this one, regardless of if it calls parent or not.
 This functions in a similar way to the `final` keyword in some languages.
+
+### `final` vars
+
+Use the above definition of FINAL_VAR to declare vars as `final`, `var/final/foo` such that overriding their value isn't permitted by types that inherit it.
+```
+/a/type
+  FINAL_VAR/foo = somevalue
+```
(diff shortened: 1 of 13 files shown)


src/dreamchecker/README.md
```diff
@@ -52,9 +52,11 @@ be enabled:
 #ifdef SPACEMAN_DMM
 	#define RETURN_TYPE(X) set SpacemanDMM_return_type = X
 	#define SHOULD_CALL_PARENT(X) set SpacemanDMM_should_call_parent = X
+  #define FINAL_VAR(X) var/final/X
 #else
 	#define RETURN_TYPE(X)
 	#define SHOULD_CALL_PARENT(X)
+  #define FINAL_VAR(X) var/X
```
SpaceManiac: Do these macros work as written with types, e.g. `FINAL_VAR(datum/reagents/reagents)`?

Maybe something like `#define FINAL_VAR var/final` or `var` then you can write `FINAL_VAR/whatever` would work better. I want to get the ergonomics on this right as far as both typed and untyped vars go.

spookydonut: They get detected the same way global/static/tmp etc get detected by the parser, so yes

spookydonut: `FINAL_VAR/whatever` might work better yeah

src/dreammaker/objtree.rs
```diff
@@ -890,12 +909,14 @@ impl ObjectTree {
             is_static,
             is_const,
             is_tmp,
+            is_final,
             type_path,
         };
         var_type.suffix(&suffix);
 
-        let symbols = &mut self.symbols;
+        let symbolid = self.symbols.allocate();
```
SpaceManiac: Old code allocates a symbol ID only if `is_declaration` is true and there wasn't already an entry, now it's unconditional. Seems like an undesirable change to me.

src/dreammaker/objtree.rs
```diff
@@ -1033,6 +1055,55 @@ impl ObjectTree {
         self.register_proc(context, location, parent, proc_name, declaration, parameters, code)
     }
 
+    pub fn check_var_defs(&self, context: &Context) {
```
SpaceManiac: Is this using any private info or can it go in `dreamchecker/lib.rs` easily?

src/dreammaker/objtree.rs
```diff
@@ -71,6 +71,7 @@ pub struct VarValue {
 #[derive(Debug, Clone)]
 pub struct TypeVar {
     pub value: VarValue,
+    pub location: Location,
```
SpaceManiac: Is this necessary? `.value.location` should be the same.

src/dreamchecker/lib.rs
```diff
@@ -492,6 +493,58 @@ fn error<S: Into<String>>(location: Location, desc: S) -> DMError {
...
+            else { continue });
+
+        guard!(let Some(typeref) = objtree.find(path)
+            else { continue });
+
+        for parent_typeref in typeref.iter_parent_types() {
+            if parent_typeref.is_root() {
+                break;
+            }
+            guard!(let parent = parent_typeref.get()
```
SpaceManiac: this guard isn't actually protecting anything

SpaceManiac:
Added some touchups on top of this
https://github.com/SpaceManiac/SpacemanDMM/compare/2b8598b...d490e3a

PR #119, opened 2019-12-03
Throw a lint error when someone does ++ or -- on non numeric types

### LINKED ISSUE #101: Throw a lint error when someone does ++ or -- on non numeric types
the default implementation on any thing that isn't a number replaces them with 1 or -1 and then all the code following will runtime.

Theoretically you can override these with proc/operator++ and proc/operator-- so maybe you can check if they have an override somewhere on their type path if possible.

Cool beans

spookydonut:
Fixes #101

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -16,7 +16,7 @@ use type_expr::TypeExpr;
 // ----------------------------------------------------------------------------
 // Helper structures
 
-#[derive(Debug, Clone)]
+#[derive(Debug, PartialEq, Clone)]
 pub enum StaticType<'o> {
     None,
     Type(TypeRef<'o>),
@@ -737,7 +737,7 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
                     ty = self.visit_follow(each.location, ty, &each.elem);
                 }
                 for each in unary.iter().rev() {
-                    ty = self.visit_unary(ty, each);
+                    ty = self.visit_unary(ty, each, location);
                 }
                 ty
             },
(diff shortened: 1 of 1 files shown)


spookydonut:
doesnt detect operator overloads yet

spookydonut:
> 
> 
> doesnt detect operator overloads yet

it does now

src/dreamchecker/lib.rs
```diff
@@ -1043,17 +1043,37 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
...
+        error(location, format!("attempted {} on a {}", crement, typeerror))
+            .register(self.context);
+        return Analysis::empty()
+    }
+
+    fn visit_unary(&mut self, rhs: Analysis<'o>, op: &UnaryOp, location: Location) -> Analysis<'o> {
         match op {
             // !x just evaluates the "truthiness" of x and negates it, returning 1 or 0
             UnaryOp::Not => Analysis::from(assumption_set![Assumption::IsNum(true)]),
+            UnaryOp::PreIncr | UnaryOp::PostIncr => { return self.check_crement(rhs, location, "operator++"); },
```
SpaceManiac: You can write `=> self.check_crement(...)` rather than having to specifically put `return` here.

src/dreamchecker/lib.rs
```diff
@@ -1043,17 +1043,37 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
...
+    // crement = increment/decrement
+    fn check_crement(&mut self, rhs: Analysis<'o>, location: Location, crement: &str) -> Analysis<'o> {
+        let typeerror;
+        match rhs.static_ty {
+            StaticType::None => {
+                return Analysis::empty()
+            },
+            StaticType::Type(typeref) => {
+                if let Some(proc) = typeref.get_proc(crement) {
+                    return Analysis::empty()
```
SpaceManiac: It would be nice if this called `visit_proc` for consistency

spookydonut: oh good point!

SpaceManiac: - Where is the `null` argument coming from?
- What does `x++` evaluate to if `operator++` is overloaded? Is `operator++`'s return value really thrown away?
- `exact` should probably be `false` because a subtype's ++ might be called.

You can fix the compile error with `let args = [...];` and `..., &args, ...`.

PR #121, opened 2019-12-06
Add warnings for ambiguous/incorrect BinaryOp::In

spookydonut:
```
/mob/proc/bar()
	var/foo = 1
	var/list/L = list()
	if((1<<0) in L) // 1
		return
		
	if(foo && foo in L) // 2
		return
		
	if((foo && foo) in L) // 3
		return

	if(!foo in L) // 4
		return
		
	if(!(foo in L)) // 5
		return

	if(foo = 2 in L) // 6
		return
		
	if(foo ? 1 : 2 in L) // 7
		return
```
detects 2, 4, 6 but not 7 due to a precedence parsing issue (#122)
```
test.dm, line 7, column 2:
warning: Found && on left side of an `in`
- add parentheses around the 'in' expression, a && (b in c)

test.dm, line 13, column 2:
warning: Found a unary ! on left side of an `in`
- add parentheses around the 'in' expression, !(a in b)

test.dm, line 19, column 2:
warning: Found assignment on left side of an `in`
```

### DIFF
--- a/src/dreammaker/parser.rs
+++ b/src/dreammaker/parser.rs
@@ -1728,11 +1728,11 @@ where
 
         // This has the effect of stripping unnecessary parentheses, which
         // simplifies later logic.
-        if unary_ops.is_empty() && follow.is_empty() {
+        /*if unary_ops.is_empty() && follow.is_empty() {
             if let Term::Expr(expr) = term.elem {
                 return success(*expr);
             }
-        }
+        }*/
 
         success(Expression::Base {
             unary: unary_ops,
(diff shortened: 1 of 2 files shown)


src/dreammaker/parser.rs
```diff
@@ -1728,11 +1728,11 @@ where
 
         // This has the effect of stripping unnecessary parentheses, which
         // simplifies later logic.
-        if unary_ops.is_empty() && follow.is_empty() {
+        /*if unary_ops.is_empty() && follow.is_empty() {
```
SpaceManiac: It needs to be investigated whether this has any ill side-effects.

spookydonut: As far as I can tell, no.

SpaceManiac: Can you test the parse of `/datum/var/x = new /datum((k = 1))` before and after this change?

spookydonut:
```

/datum/var/x = new /datum((k = 1))
```
commented out:
```

test.dm, line 2, column 14:
error: non-constant assignment
```
uncommented:
none

PR #123, opened 2019-12-06
Add basic unreachable code detection

spookydonut:
Has some simple evaluation

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -791,7 +791,6 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
                 return term // stop evaluating
             }
             let state = self.visit_statement(stmt.location, &stmt.elem);
-            //println!("{:#?} {:#?}", stmt, state);
             term.merge(state);
         }
         return term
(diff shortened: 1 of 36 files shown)


spookydonut:
This needs work, I'm making progress but I just realised a mistake I made in truthy evaluation and what I can do to fix it

src/dreamchecker/lib.rs
```diff
@@ -545,6 +545,91 @@ pub fn check_var_defs(objtree: &ObjectTree, context: &Context) {
...
+        self.continues = false;
+        self.breaks = false;
+    }
+
+    pub fn merge(&mut self, other: ControlFlow) {
+        if !self.fuzzy && other.returns {
+            self.returns = true;
+        }
+        if other.fuzzy {
+            self.returns = false;
```
Cyberboss: ?

spookydonut: loops basically

spookydonut:
I'm pretty happy with the state of this

src/dreamchecker/lib.rs
spookydonut: you can't (yet) determine if a loop ran, so it's "fuzzy" if it can actually return and terminate the proc leaving the code below the loop unreachable.

PR #124, opened 2019-12-07
Add lints for private/protected vars and procs

spookydonut:
See unit tests for proof

from README
```md
### Private / Protected procs

Use `set SpacemanDMM_private_proc = 1` and 
`set SpacemanDMM_protected_proc = 1` to set procs as private and protected respectively.

* Private procs can only be called by things of exactly the same type
* Protected procs can only be call by things of the same type or subtypes

Additionally, Private procs cannot be overridden.

### Private / Protected vars

Use the above definitions of VAR_PRIVATE and VAR_PROTECTED to declare vars as 
`SpacemanDMM_private`/`SpacemanDMM_protected`.
These function the same way as the proc versions.
```


spookydonut:
dont merge this theres some debate about how it should work

spookydonut:
This is ready

spookydonut:
Now with unit tests

PR #142, opened 2020-01-08
Revert "Adds /client/SoundQuery to dreammaker"

yoyobatty:
Reverts SpaceManiac/SpacemanDMM#139

Duplicate proc: https://github.com/SpaceManiac/SpacemanDMM/blob/1bb11a4f24c779ef5ec8e0f1803063887d10d11c/src/dreammaker/builtins.rs#L964


SpaceManiac:
should have known morty would do me dirty

SpaceManiac:
I guess I should add tests for this

PR #146, opened 2020-01-15
Config option to ignore directories/files

spookydonut:
as requested.


SpaceManiac:
I'm not interested in maintaining support for this unless it's proven necessary in practice and not just in theory.

PR #147, opened 2020-01-15
Readds override precedes definition warning

spookydonut:
now with configuration option


src/dreammaker/objtree.rs
```diff
@@ -994,7 +994,18 @@ impl ObjectTree {
         // implementing, so abusers of it will have to put up with some
```
SpaceManiac: You may want to update this TODO comment

PR #149, opened 2020-01-15
Add option for dmdoc to always use typepath names

### LINKED ISSUE #94: Option to disable using obj var name for docs
The object var name is normally related to the display of the object in game, not something the developer would recognise immediately, I'd rather use the type paths exactly.

A good example is /area/, which shows up as "Space" not "area", which confuses the hell out of me when browsing the docs.

spookydonut:
Closes #94

### DIFF
--- a/src/dreammaker/config.rs
+++ b/src/dreammaker/config.rs
@@ -16,6 +16,7 @@ pub struct Config {
     display: WarningDisplay,
     pub langserver: Langserver,
     diagnostics: HashMap<String, WarningLevel>,
+    dmdoc: DMDoc,
 }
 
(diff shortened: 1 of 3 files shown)


src/dreammaker/config.rs
```diff
@@ -16,6 +16,7 @@ pub struct Config {
     display: WarningDisplay,
     pub langserver: Langserver,
     diagnostics: HashMap<String, WarningLevel>,
+    dmdoc: DMDoc,
```
SpaceManiac: should be fine to just make `dmdoc` pub and access the field directly

PR #150, opened 2020-01-15
Add off-by-default lint to forbid relative pathing

### LINKED ISSUE #148: Warn about relative pathing
Relative pathing is annoying. Please add a config option to warn about files that contain relative pathing.

spookydonut:
Closes #148

This detects type definitions without a leading `/` as relative though.

### DIFF
--- a/src/dreammaker/config.rs
+++ b/src/dreammaker/config.rs
@@ -16,6 +16,7 @@ pub struct Config {
     display: WarningDisplay,
     pub langserver: Langserver,
     diagnostics: HashMap<String, WarningLevel>,
+    code_standards: CodeStandards,
 }
 
 #[derive(Deserialize, Default, Debug, Clone)]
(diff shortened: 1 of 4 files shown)


src/dreammaker/config.rs
```diff
@@ -29,6 +30,14 @@ pub struct Langserver {
     pub dreamchecker: bool,
 }
 
+#[derive(Deserialize, Default, Debug, Clone)]
+pub struct CodeStandards {
+    #[serde(default)]
```
SpaceManiac: This probably belongs on the struct rather than on only one of the two field

src/dreammaker/objtree.rs
```diff
@@ -1008,7 +1014,7 @@ impl ObjectTree {
             elems.len() + 1,
             Default::default(),
             Default::default(),
-        )
+        ).and(Ok(()))
```
SpaceManiac: I would rather see `(...)?; Ok(())` than this use of `.and()`, ditto below

src/dreammaker/config.rs
```diff
@@ -74,6 +83,10 @@ impl Config {
     pub fn registerable_error(&self, error: &DMError) -> bool {
         self.display.error_level.applies_to(error.severity())
     }
+
+    pub fn code_standards(&self) -> &CodeStandards {
+        &self.code_standards
```
SpaceManiac: It's probably fine if this is a `pub` field but it's not that important either way

PR #154, opened 2020-01-17
Warn about invalid filter() kwargs

spookydonut:
Removes a TODO and leaves a new one.


spookydonut:
Do not merge, this is not working correctly

spookydonut:
Fixed it

PR #156, opened 2020-01-19
Add targeted dreamchecker unit testing

spookydonut:


src/dreammaker/error.rs
```diff
@@ -82,6 +82,14 @@ impl FileList {
 }
 
 impl Context {
+    pub fn unit_test_default() -> Context {
```
SpaceManiac: How is this different from the Default impl?

PR #158, opened 2020-01-21
Add check_var_defs to dreamchecker::run

spookydonut:
Add comments and move check_var_defs to ObjTree

### DIFF
--- a/src/dreammaker/objtree.rs
+++ b/src/dreammaker/objtree.rs
@@ -531,6 +531,12 @@ impl<'a> ProcRef<'a> {
         self.ty.tree()
     }
 
+    /// Check if the target ProcValue is a builtin.
+    #[inline]
+    pub fn is_builtin(self) -> bool {
+        self.get().location.is_builtins()
+    }
+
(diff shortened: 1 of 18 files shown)


src/dreammaker/objtree.rs
```diff
@@ -1140,6 +1140,57 @@ impl ObjectTree {
             }
         }
     }
+
+    // ----------------------------------------------------------------------------
+    // Variable analyzer
+
+    /// Checks for var declaration issues
+    pub fn check_var_defs(&self, context: &Context) {
```
SpaceManiac: Why does this need to be here?

spookydonut: it's an examination of the object tree, feels like it belongs here more than just being a global function

SpaceManiac: "checking var defs" is a dreamchecker responsibility so it belongs in dreamchecker IMO. Rust encourages organization by modules rather than just by "most important input type".

spookydonut: fair enough

PR #159, opened 2020-01-21
Fix detection of ambig ternary in operations

### LINKED ISSUE #122: Ternary has precedence over BinaryOp::In
`(foo ? foo : 2 in L)`
byond parses this as
`((foo ? foo : 2) in L)`
but we parse it as
`(foo ? foo : (2 in L))`
```
BinaryOp {
    op: In,
    lhs: Base {
        unary: [],
        term: Spanned {
            location: Location {
                file: FileId(
                    2,
                ),
                line: 22,
                column: 15,
            },
            elem: Int(
                2,
            ),
        },
        follow: [],
    },
    rhs: Base {
        unary: [],
        term: Spanned {
            location: Location {
                file: FileId(
                    2,
                ),
                line: 22,
                column: 20,
            },
            elem: Ident(
                "L",
            ),
        },
        follow: [],
    },
}
```

spookydonut:
Related #122

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -907,6 +907,14 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
                 // TODO: be sensible
                 self.visit_expression(location, cond, None);
                 let ty = self.visit_expression(location, if_, type_hint);
+                if let Expression::BinaryOp{ op: BinaryOp::In, lhs: _, rhs: _ } = **else_ {
+                    error(location, format!("ambiguous `in` on right side of a ternary"))
+                        .set_severity(Severity::Warning)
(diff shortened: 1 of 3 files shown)


SpaceManiac:
Isn't the issue that the parser produces the wrong output?

Seems like the original detection code would work if the parser was fixed.

Alternatively, checking if `?:` contains an `in` on its right-hand could just as well happen in dreamchecker as the parser.

SpaceManiac:
I think I'd rather fix the parser's output than hack the diagnostic

spookydonut:
> 
> 
> I think I'd rather fix the parser's output than hack the diagnostic

I didn't see a trivial way to do that.

SpaceManiac:
the bug can stand while we stew on a solution

PR #160, opened 2020-01-21
Filter flags checking

spookydonut:
This is sort of a template for deal with other flags of this kind, i dont think it's done the best though


spookydonut:
Travis failure is correct as it now checks flags values

PR #161, opened 2020-01-22
These defines were wrong

spookydonut:
thanks stddef.dm

### DIFF
--- a/src/dreammaker/builtins.rs
+++ b/src/dreammaker/builtins.rs
@@ -163,25 +163,18 @@ pub fn default_defines(defines: &mut DefineMap) {
         MASK_SWAP = Int(2);
 
         // rgb filter
-        FILTER_COLOR_RGB = Int(1);
-        FILTER_COLOR_HSV = Int(2);
-        FILTER_COLOR_HSL = Int(4);
-        FILTER_COLOR_HCY = Int(8);
+        FILTER_COLOR_RGB = Int(0);
+        FILTER_COLOR_HSV = Int(1);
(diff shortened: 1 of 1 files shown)


src/dreammaker/builtins.rs
```diff
@@ -163,25 +163,18 @@ pub fn default_defines(defines: &mut DefineMap) {
...
+        FILTER_COLOR_HSL = Int(2);
+        FILTER_COLOR_HCY = Int(3);
 
-        // layering
+        // layering / ray filter
         FILTER_OVERLAY = Int(1);
         FILTER_UNDERLAY = Int(2);
 
-        // ray filter
-        FLAG_OVERLAY = Int(1);
```
SpaceManiac: You should report a docs bug given that these are described in the ray filter docs:
http://www.byond.com/docs/ref/info.html#/{notes}/filters/rays

src/dreammaker/builtins.rs
```diff
@@ -163,25 +163,18 @@ pub fn default_defines(defines: &mut DefineMap) {
...
-        FLAG_UNDERLAY = Int(2);
-
-        // ripple filter
-        //WAVE_BOUND = Int(2);
-
-        // wave filter
+        // wave filter / ripple filter
         WAVE_SIDEWAYS = Int(1);
-        WAVE_BOUND = Int(2);
+        WAVE_BOUNDED = Int(2);
```
SpaceManiac: Ditto docs bug

PR #167, opened 2020-01-25
Add checking of parameter StaticType as Info diagnostic

spookydonut:
still needs some more work to check for false positives i think

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -9,6 +9,7 @@ use dm::objtree::{ObjectTree, TypeRef, ProcRef, Code};
 use dm::constants::{Constant, ConstFn};
 use dm::ast::*;
 
+use std::fmt;
 use std::collections::{BTreeMap, HashMap, HashSet};
 
 mod type_expr;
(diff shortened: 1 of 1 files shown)


src/dreamchecker/lib.rs
```diff
@@ -69,6 +81,19 @@ impl<'o> StaticType<'o> {
...
         StaticType::List { list: tree.expect("/list"), keys: Box::new(StaticType::Type(tree.expect(of))) }
     }
+
+    fn is_subtype_of(&self, other: &StaticType<'o>) -> bool {
+        match (self, other) {
+            (StaticType::Type(selftype), StaticType::Type(othertype)) => {
+                selftype.is_subtype_of(othertype.get())
+            },
+            (StaticType::List{list: selflist, keys: _}, StaticType::List{list: otherlist, keys: _}) => {
+                selflist.is_subtype_of(otherlist.get())
```
SpaceManiac: you probably want to be comparing `keys`, not `list`. `list` is always initialized to a reference to `/list`

spookydonut: suddenly this enum makes way more sense

PR #170, opened 2020-03-05
Add double parentcall detection

spookydonut:
todo: unit tests


spookydonut:
This has a lot of detections on tg

PR #173, opened 2020-04-01
Fix locations for if else arm and switch if arms

spookydonut:
These previously gave weird locations


SpaceManiac:
need to update the tests

PR #174, opened 2020-04-04
Fold sdmm parser into the repo

spookydonut:
this is useful and it'd be great to keep it updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1477,6 +1477,16 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "sdmmparser"
+version = "0.0.2"
+dependencies = [
+ "dreammaker 0.1.0",
+ "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


SpaceManiac:
Releases already include 4 different binaries. I'm not sure I want to make that number bigger.

skull132:
> 
> 
> Releases already include 4 different binaries. I'm not sure I want to make that number bigger.

IMO this is a relatively useful binary to have, as it allows SDMM to be applied to other processes relatively easily. I already have plans to automate certain wiki articles with this, and someone from CM wanted a tool like this to be used for balancing.

If binary bloat is the issue, then why not consider rolling some of these into a single binary that is switchable, ala git?

SpaceManiac:
> If binary bloat is the issue, then why not consider rolling some of these into a single binary that is switchable, ala git?

What are our main consumers?
- MapDiffBot and AA's webmap, using dmm-tools compiled from source (sometimes modified)
- Travis, using dmdoc and dreamchecker downloaded from GitHub releases
- VSC extension, using langserver downloaded from private infrastructure

That covers the four binaries that currently matter. langserver and dmm-tools are made available on GitHub releases for completeness. Where can we make this fit? If we rearranged existing binaries, what would we combine?

The main force in the opposite direction of bundling is the difference in dependency sets between the different components, which manifests in build time and in file size.

spookydonut:
strongdmm has been updated.

PR #176, opened 2020-04-07
Fixes ';' in proc args causing parse failure

### LINKED ISSUE #172: ; in proc arguments causes a fatal parse error
https://github.com/PolarisSS13/Polaris/pull/6941

Cyberboss:
Optionally allow trailing `;`s after the expression in proc arguments. Emit a warning when doing so.

Added ast_tests.

Fixes #172

### DIFF
--- a/src/dreammaker/tests/ast_tests.rs
+++ b/src/dreammaker/tests/ast_tests.rs
@@ -1,0 +1,34 @@
+extern crate dreammaker as dm;
+
+use dm::*;
+use dm::preprocessor::Preprocessor;
+use dm::objtree::ObjectTree;
+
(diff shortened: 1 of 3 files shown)


src/dreammaker/tests/ast_tests.rs
```diff
@@ -0,0 +1,34 @@
...
+    f(context, _tree)
+}
+
+#[test]
+fn check_semicolon_in_proc_parameters() {
+    with_code("
+#define SOMESHIT 0x01;
+#define SUMFUCK \"asdf\" as text;
+
+/proc/fuck(foo = SOMESHIT, bar = SUMFUCK, anotherarg = 1)
```
SpaceManiac, line 25: SFW naming please

PR #197, opened 2020-06-29
fix precedence for bit operations

### LINKED ISSUE #196: bad precedence for bit ops
In DM, `1 | 6 & 2` is parsed as `1 | (6 & 2)` and returns 3.
Instead, we parse it as `(1 | 6) & 2`, as all bit ops are treated as having the same precedence.
According to ref, precedence should be `&` > `^` > `|`

mloc:
Fixes #196


mloc:
Travis timed out

SpaceManiac:
> Travis timed out

Cache config on this repo is not correct, so that happens when the cache gets too big. I did the temporary fix

PR #201, opened 2020-07-19
Populate errortype on more errors.

Leshana:
- Added unreachable_code, proc_has_no_parent, and loop_condition_determinate error types.
- Added one use of control_condition_static where it was missing.
- Added unit tests covering some of the error conditions I looked at while adding these.

This is convenient for me when working on older codebases where I want to categorize and prioritize the types of linting errors to fix.  In addition to the ones I wanted, @spookydonut 's code inspired a few additional ones to add.
Plus I figure more unit test coverage is always helpful.

### DIFF
--- a/CONFIGURING.md
+++ b/CONFIGURING.md
@@ -43,9 +43,12 @@ Raised by DreamChecker:
 * `no_typehint_implicit_new` - Raised on the use of `new` where no typehint is avaliable
 * `field_access_static_type` - Raised on using `.field_name` on a variable with no typehint
 * `proc_call_static_type` - Raised on using `.proc_name()` on a variable with no typehint
+* `proc_has_no_parent` - Raised on calling `..()` in a proc with no parent.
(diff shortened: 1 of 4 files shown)


spookydonut:
@Leshana those errors are intentionally not able to be disabled.

SpaceManiac:
> those errors are intentionally not able to be disabled.

@spookydonut Could you elaborate on the reasoning for this?

spookydonut:
> 
> 
> > those errors are intentionally not able to be disabled.
> 
> @spookydonut Could you elaborate on the reasoning for this?

@SpaceManiac If people don't want errors raised by lints they should... not use those lints?

spookydonut:
> 
> 
> > > those errors are intentionally not able to be disabled.
> > 
> > 
> > @spookydonut Could you elaborate on the reasoning for this?
> 
> @SpaceManiac If people don't want errors raised by lints they should... not use those lints?

Oh wait, nvm there's no reason not to allow these to be configged

PR #206, opened 2020-08-14
Adds missing filter flags for the outline filter.

### LINKED ISSUE #205: DM-Langserver does not know about builtin Filter flags
BYOND has a large number of built in flags, and I assume that the fairly new BYOND visual filter flags have yet to get added to a list of built in flags. The result is that using one of those flags results in an erroneous unknown variable error, when it compiles just fine. The flag in question is `OUTLINE_SHARP` for the outline filter, [as seen in this picture](https://i.imgur.com/OU6Cdhe.png).

After doing some research, the built in flag definitions seem to be [located here](https://github.com/SpaceManiac/SpacemanDMM/blob/master/src/dreammaker/builtins.rs#L164) so I'm not sure if it's as simple as just adding them so the language server stops thinking it's an undefined variable. If it actually is just that simple then I can try to fix it, and #195 seems to suggest that it would be an easy fix, hopefully.

Neerti:
Fixes #205 
Tried to fix the issue I found last night in the same way that #165 added other filter flags, but that PR seemed to have forgotten to add the two outline flags.

I tested this locally and the error doesn't come up now so it does work.

Hopefully I did everything right, let me know if there's any issues.

### DIFF
--- a/src/dreammaker/builtins.rs
+++ b/src/dreammaker/builtins.rs
@@ -328,6 +328,10 @@ pub fn register_builtins(tree: &mut ObjectTree) -> Result<(), DMError> {
         var/const/WAVE_SIDEWAYS = int!(1);
         var/const/WAVE_BOUNDED = int!(2);
 
+        // outline filter
+        var/const/OUTLINE_SHARP = int!(1);
+        var/const/OUTLINE_SQUARE = int!(2);
+
         // global procs
         proc/abs(A);
         proc/addtext(Arg1, Arg2/*, ...*/);


ZeWaka:
You'll need to add stuff here I think:
https://github.com/SpaceManiac/SpacemanDMM/blob/ec16f1202822f3b7e4f7c56151830b7a1a8e7cc1/src/dreammaker/ast.rs#L918-L942

Neerti:
When I look on the latest version on the master branch, that block of code appeared to already have the outline stuff there.
https://github.com/SpaceManiac/SpacemanDMM/blob/ac037752c930030a46f4ddcb7e39b199246d19d1/src/dreammaker/ast.rs#L1143
https://github.com/SpaceManiac/SpacemanDMM/blob/ac037752c930030a46f4ddcb7e39b199246d19d1/src/dreammaker/ast.rs#L1156

PR #211, opened 2020-09-08
Remove error if a proc has both should_not_sleep and set waitfor = 0

MrStonedOne:
See https://github.com/tgstation/tgstation/pull/53553#discussion_r485090104 for why this logic is invalid.

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -632,13 +632,6 @@ impl<'o> AnalyzeObjectTree<'o> {
 
     pub fn check_proc_call_tree(&mut self) {
         for (procref, &(_, location)) in self.must_not_sleep.directive.iter() {
-            if let Some(_) = self.waitfor_procs.get(&procref) {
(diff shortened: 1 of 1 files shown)


optimumtact:
There should be  a different lint define for this case, rather than allowing set waitfor when people expect no sleeping.

MrStonedOne:
But set waitfor doesn't sleep!

It just guarantees the caller will not be locked up, the same thing should_not_sleep tries, but fails to guarantee.

the only time set waitfor = false would kick in, is in a sleep that `should_not_sleep` doesn't catch. 

And because byond is not a statically typed language, should_not_sleep can not catch every case where something can sleep, trying to is as mathematically impossible as the halting problem, you can get very close, but not perfect detection.

It makes no sense to force disabling of set waitfor = false to use `should_not_sleep`, that seems like the kind of thing where, "We are not the only people using this", so we shouldn't be telling them if they can or can not use both at once.

SpaceManiac (approved):
`should_not_sleep` and `set waitfor = FALSE` are static and dynamic ways to (attempt to) accomplish the same goal of not blocking the caller. They are not in conflict and warning on their combination is unnecessary.

PR #213, opened 2020-10-04
Implements multiple suggested clippy lints

ZeWaka:
These were the lints mostly:
* match_like_matches_macro
* mem_replace_with_default
* match_ref_pats
* single_char_push_str
* len_zero

There were also a few unnecessary return lints that popped up, but I'm not sure if you want to remove those.

### DIFF
--- a/src/dreammaker/ast.rs
+++ b/src/dreammaker/ast.rs
@@ -553,11 +553,11 @@ pub enum Term {
 impl Term {
     pub fn is_static(&self) -> bool {
         matches!(self,
-            Term::Null      |
-            Term::Int(_)    |
-            Term::Float(_)  |
-            Term::String(_) |
-            Term::Prefab(_)
+            Term::Null
+            | Term::Int(_)
+            | Term::Float(_)
+            | Term::String(_)
+            | Term::Prefab(_)
         )
     }
 
--- a/src/dreammaker/ast.rs
+++ b/src/dreammaker/ast.rs
@@ -552,14 +552,13 @@ pub enum Term {
 
 impl Term {
     pub fn is_static(&self) -> bool {
-        return match self {
-            Term::Null |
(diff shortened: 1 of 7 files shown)


src/dreammaker/ast.rs
```diff
@@ -552,14 +552,13 @@ pub enum Term {
 
 impl Term {
     pub fn is_static(&self) -> bool {
-        return match self {
-            Term::Null |
-            Term::Int(_) |
-            Term::Float(_) |
+        matches!(self,
+            Term::Null      |
```
SpaceManiac, line 556: not a fan of the alignment being used here and on the other `matches!`, not sure if that was you or the tool though

SpaceManiac:
> unnecessary return lints

probably fine to fix, at least some of them are leftovers from refactors

src/dreammaker/ast.rs
ZeWaka: It was me, how would you prefer it to look like?

spookydonut (approved):

src/dreammaker/ast.rs
SpaceManiac: `|` either at the end of the line without spaces, or at the beginning of the following line...

src/dreammaker/dmi.rs
```diff
@@ -51,13 +51,10 @@ impl Dir {
...
     pub fn is_diagonal(self) -> bool {
-        match self {
-            Dir::North |
-            Dir::South |
-            Dir::East |
-            Dir::West => false,
-            _ => true
-        }
+        !matches!(self, Dir::North  |
+                        Dir::South  |
```
SpaceManiac, line 55: and preferably a single indent here rather than hanging alignment

SpaceManiac (approved):

PR #216, opened 2020-10-09
Fix "Should not sleep" to catch world.Export and world.Import

### LINKED ISSUE #208: "Should not sleep" does not catch world.Export or world.Import

spookydonut:
Fixes #208


src/dreamchecker/lib.rs
```diff
@@ -1702,10 +1702,13 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
...
 
             Term::Call(unscoped_name, args) => {
-                if unscoped_name == "sleep" || unscoped_name == "alert" || unscoped_name == "shell" || unscoped_name == "winexists" || unscoped_name == "winget" {
-                    if self.inside_newcontext == 0 {
+                if matches!(unscoped_name.as_str(),
+                    "sleep"
+                    | "alert"
+                    | "shell"
+                    | "winexists"
+                    | "winget" if self.inside_newcontext == 0) {
```
SpaceManiac, line 1710: style nit: the `if` guard doesn't depend on the pattern, so it should probably go outside the `matches!()` invocation

SpaceManiac (changes requested):
also test doesn't pass

src/dreamchecker/lib.rs
spookydonut: true

PR #217, opened 2020-10-10
Add can be builtin

### LINKED ISSUE #202: would be nice to be able to use stuff like SHOULD_NOT_SLEEP on builtin procs
say you got atom/movable/Move() but I can't do that since I'm not defining it on the "base" and I can't just redefine it because it's.. already defined

but yeah would be nice to have

spookydonut:
Fixes #202


SpaceManiac:
Do we need to afford special status to builtins here? Can we simply say these lint settings "must be on the declaration" or not?

spookydonut:
> 
> 
> Do we need to afford special status to builtins here? Can we simply say these lint settings "must be on the declaration" or not?

@SpaceManiac maybe just allow compatible lints on built-ins regardless?

SpaceManiac:
@spookydonut why does "built-in-ness" matter at all? Are there lints that don't work if they go on overrides rather than `/proc/` declarations?

PR #228, opened 2020-11-05
Add warning for `return x` in spawn context

pali6:
As far as I know doing `return something` in a `spawn` context is always equivalent to just returning without a value and as such it is probably a mistake so let's warn people about it.

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -1281,6 +1281,11 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
             },
             Statement::Return(Some(expr)) => {
                 // TODO: factor in the previous return type if there was one
+                if self.inside_newcontext > 0 {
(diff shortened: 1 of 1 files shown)


src/dreamchecker/lib.rs
```diff
@@ -1281,6 +1281,11 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
             },
             Statement::Return(Some(expr)) => {
                 // TODO: factor in the previous return type if there was one
+                if self.inside_newcontext > 0 {
```
Cyberboss, line 1284: IIRC isn't this also set in waitfor = 0 procs?

pali6: I don't think so, the only place that changes `inside_newcontext` to non-zero values seems to be the `Statement::Spawn` match arm.

PR #229, opened 2020-11-05
Fixes var declaration location being wrong

### LINKED ISSUE #219: VarDeclaration have incorrect locations
```
/datum
    var/bar

/datum/proc/foo()
```

bar's location is incorrectly reported at what i assume is the next token, foo()

pali6:
Fixes #219 .
It looks like this is caused by two separate issues. One is that `put_back` doesn't update `location` properly and the other is that in declarations with assignment the location used for var declaration is taken after parsing the `=`. This hopefully fixes both.

I checked that var declaration locations now seem to be correct, I just hope nothing relies on `put_back` not fixing location. Yell at me if I got something completely wrong here.

### DIFF
--- a/src/dreammaker/parser.rs
+++ b/src/dreammaker/parser.rs
@@ -289,6 +289,7 @@ pub struct Parser<'ctx, 'an, 'inp> {
     possible_indentation_error: bool,
     next: Option<Token>,
     location: Location,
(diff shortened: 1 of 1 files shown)


pali6:
Welp, it seems like a bunch of stuff sorta relies on the put_back behaviour and it seems a bit of a mess. Will reopen if/when I fix.

PR #230, opened 2020-11-26
Auxtools debugger

willox:
This adds support for a debug server written in auxtools (currently located at https://github.com/willox/auxtools.)

The dependency is similar to extools, where SpacemanDMM's `auxtools_types.rs` has to be up-to-date with the `server_types.rs` file located in whichever version of debug server is used. In the future this could change to be some shared dependency, or maybe the debug server could just be moved into SpacemanDMM.

There's a bunch of repeated code in `mod.rs` where there's match statements where one branch is for the extools client and one branch is for auxtools client. It's a bit iffy, but they all have minor differences and wouldn't be super easy to merge.

I accidentally ran a `cargo fmt` on the files I was working with at some point, so there's a few formatting changes about. I don't think it's too much to read over.

I haven't edited any documentation yet, so here's how it works:
In your DM project:
```dm
// Currently needed for auxtools' error reporting. TG code already has this defined.
/proc/stack_trace(msg)
	CRASH(msg)

/proc/enable_debugging(mode, port)
	CRASH("auxtools not loaded")

/world/New()
	var/debug_server = world.GetConfig("env", "AUXTOOLS_DEBUG_DLL")
	if (debug_server)
		call(debug_server, "auxtools_init")()
		enable_debugging()
	. = ..()

/world/Del()
	var/debug_server = world.GetConfig("env", "AUXTOOLS_DEBUG_DLL")
	if (debug_server)
		call(debug_server, "auxtools_shutdown")()
	. = ..()
```

In your project's SpacemanDMM.toml
```toml
[debugger]
engine = "auxtools"
```

The extension doesn't have a way to override the DLL being used (and I don't think it should), so if you're testing stuff I suggest you set the env vars to something like below and use the attach mode:
```
AUXTOOLS_DEBUG_DLL=path_to_your_build
AUXTOOLS_DEBUG_MODE=BLOCK
```

LMK what needs changing

### DIFF
--- a/src/langserver/Cargo.toml
+++ b/src/langserver/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Tad Hardesty <tad@platymuus.com>"]
 edition = "2018"
 
 [[bin]]
-name = "dm-langserver"
+name = "dm_langserver"
 path = "main.rs"
 
 [dependencies]
--- a/src/langserver/Cargo.toml
+++ b/src/langserver/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Tad Hardesty <tad@platymuus.com>"]
 edition = "2018"
 
 [[bin]]
-name = "dm_langserver"
+name = "dm-langserver"
 path = "main.rs"
 
 [dependencies]
--- a/src/langserver/Cargo.toml
+++ b/src/langserver/Cargo.toml
@@ -13,6 +13,7 @@ url = "2.1.0"
 serde = "1.0.27"
 serde_json = "1.0.10"
 serde_derive = "1.0.27"
+bincode = "1.3.1"
 jsonrpc-core = "14.0.3"
 lsp-types = "0.80.0"
 dreammaker = { path = "../dreammaker" }
--- a/src/langserver/debugger/auxtools.rs
+++ b/src/langserver/debugger/auxtools.rs
@@ -13,212 +13,226 @@ use std::{
 use super::dap_types;
 use super::SequenceNumber;
 
+enum StreamState {
+	// The client is waiting for a Stream to be sent from the thread
+	Waiting(mpsc::Receiver<TcpStream>),
+
+	Connected(TcpStream),
+
+	// The server has finished being used
+	Disconnected,
+}
+
 pub struct Auxtools {
+    seq: Arc<SequenceNumber>,
     responses: mpsc::Receiver<Response>,
     _thread: JoinHandle<()>,
-    stream: TcpStream,
+    stream: StreamState,
     last_error: Arc<RwLock<String>>,
 }
 
 pub struct AuxtoolsThread {
     seq: Arc<SequenceNumber>,
     responses: mpsc::Sender<Response>,
-    stream: TcpStream,
     last_error: Arc<RwLock<String>>,
 }
 
 impl Auxtools {
-    pub fn new(seq: Arc<SequenceNumber>, port: Option<u16>) -> std::io::Result<Self> {
+    pub fn connect(seq: Arc<SequenceNumber>, port: Option<u16>) -> std::io::Result<Self> {
         let addr: SocketAddr = (Ipv4Addr::LOCALHOST, port.unwrap_or(DEFAULT_PORT)).into();
         let (responses_sender, responses_receiver) = mpsc::channel();
-
+        let last_error = Arc::new(RwLock::new("".to_owned()));
         let stream = TcpStream::connect_timeout(&addr, std::time::Duration::from_secs(5))?;
 
         // Look at this little trouble-maker right here
         seq.issue_event(dap_types::InitializedEvent);
 
-        let last_error = Arc::new(RwLock::new("".to_owned()));
-
-        let thread = AuxtoolsThread {
-            seq,
-            responses: responses_sender,
-            stream: stream.try_clone().unwrap(),
-            last_error: last_error.clone(),
+        let thread = {
+            let seq = seq.clone();
+            let last_error = last_error.clone();
+            let stream = stream.try_clone().unwrap();
+            thread::spawn(move || {
+                AuxtoolsThread {
+                    seq,
+                    responses: responses_sender,
+                    last_error: last_error,
+                }.run(stream);
+            })
         };
 
         Ok(Auxtools {
+            seq,
             responses: responses_receiver,
-            _thread: thread.start_thread(),
-            stream,
+            _thread: thread,
+            stream: StreamState::Connected(stream),
             last_error,
         })
     }
 
-    fn read_response(&mut self) -> Result<Response, mpsc::RecvTimeoutError> {
-        self.responses
-            .recv_timeout(std::time::Duration::from_secs(5))
+    fn read_response_or_disconnect(&mut self) -> Result<Response, Box<dyn std::error::Error>> {
+        match self.responses.recv_timeout(std::time::Duration::from_secs(5)) {
+            Ok(response) => Ok(response),
+            Err(_) => {
+                self.disconnect();
+                Err(Box::new(super::GenericError("timed out waiting for response")))
+            }
+        }
     }
 
-    fn send(&mut self, request: Request) -> Result<(), ()> {
-        let data = bincode::serialize(&request).unwrap();
-        self.stream.write_all(&(data.len() as u32).to_le_bytes()).unwrap();
-        self.stream.write_all(&data[..]).unwrap();
-        self.stream.flush().unwrap();
+    fn send(&mut self, request: Request) -> Result<(), Box<dyn std::error::Error>> {
+        if let StreamState::Waiting(recv) = &self.stream {
+            if let Ok(stream) = recv.try_recv() {
+                self.stream = StreamState::Connected(stream);
+            }
+        }
+
+        match &mut self.stream {
+            StreamState::Connected(stream) => {
+                let data = bincode::serialize(&request)?;
+                stream.write_all(&(data.len() as u32).to_le_bytes())?;
+                stream.write_all(&data[..])?;
+                stream.flush()?;
+                Ok(())
+            }
+
+            _ => {
+                // Success if not connected (kinda dumb)
+                Ok(())
+            }
+        }
+    }
+
+    fn send_or_disconnect(&mut self, request: Request) -> Result<(), Box<dyn std::error::Error>>  {
+        if let Err(e) = self.send(request) {
+            self.disconnect();
+            return Err(e);
+        }
+
         Ok(())
     }
 
     pub fn disconnect(&mut self) {
-        self.send(Request::Disconnect).unwrap();
+        debug_output!(in self.seq, "[auxtools] disconnecting");
+        let _ = self.send(Request::Disconnect);
+
+        if let StreamState::Connected(stream) = &self.stream {
+            let _ = stream.shutdown(std::net::Shutdown::Both);
+        }
+
+        self.stream = StreamState::Disconnected;
     }
 
-    pub fn get_line_number(&mut self, path: &str, override_id: u32, offset: u32) -> Option<u32> {
-        self.send(Request::LineNumber {
-                proc: ProcRef {
-                    path: path.to_owned(),
-                    override_id,
-                },
-                offset,
-            })
-            .unwrap();
+    pub fn get_line_number(&mut self, path: &str, override_id: u32, offset: u32) -> Result<Option<u32>, Box<dyn std::error::Error>> {
+        self.send_or_disconnect(Request::LineNumber {
+            proc: ProcRef {
+                path: path.to_owned(),
+                override_id,
+            },
+            offset,
+        })?;
 
-        match self.read_response() {
-            Ok(response) => {
-                match response {
-                    Response::LineNumber { line } => line,
-
-                    // TODO: disconnect
-                    _ => panic!("received wrong response"),
-                }
-            }
-
-            // TODO: disconnect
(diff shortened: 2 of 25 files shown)


src/langserver/Cargo.toml
```diff
@@ -5,14 +5,15 @@ authors = ["Tad Hardesty <tad@platymuus.com>"]
 edition = "2018"
 
 [[bin]]
-name = "dm-langserver"
+name = "dm_langserver"
```
willox, line 8: this was changed to avoid a rust bug where symbols are not generated (at least on Windows with MSVC) if a dash is in the name

SpaceManiac: is there an upstream ticket?

SpaceManiac:
> The extension doesn't have a way to override the DLL being used (and I don't think it should)

The extension currently has a hidden setting (not exposed in the UI) to override the extools DLL for testing purposes, would the same work here?

willox:
Yeah, I haven't added the equivalent for auxtools because I don't see it actually helping me. There was maybe a reason for people to use it before when developing their own extensions onto extools but with auxtools we can have multiple separate libraries loaded at once.

We can add it, but changing that option just seems more complicated than setting the env-var.

src/langserver/Cargo.toml
willox: for this specific issue idk, it was a problem before but got 'fixed'.
https://github.com/rust-lang/cargo/pull/8123
perhaps it was only fixed when there was no `[[bin]].name` override.

src/langserver/debugger/mod.rs
```diff
@@ -57,37 +66,49 @@ pub fn start_server(dreamseeker_exe: String, db: DebugDatabaseBuilder) -> std::i
         .spawn(move || {
             let (stream, _) = listener.accept().unwrap();
             drop(listener);
-
+            let env = dm::detect_environment_default()
+                .expect("error detecting .dme")
+                .expect("no .dme found");
+            let ctx = dm::Context::default();
+            ctx.autodetect_config(&env);
```
SpaceManiac, line 73: The `dm::Context` or config should be passed in from the language server here, rather than going through `detect_environment`.

src/langserver/debugger/auxtools_types.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use serde::{Deserialize, Serialize};
+
+#[allow(dead_code)]
+pub const DEFAULT_PORT: u16 = 2448;
+
+// Message from client -> server
+#[derive(Serialize, Deserialize, Debug)]
+pub enum Request {
+	Disconnect,
```
SpaceManiac, line 11: File should be 4-spaced, not tabbed

src/langserver/debugger/auxtools.rs
```diff
@@ -0,0 +1,432 @@
...
+    net::TcpStream,
+    sync::{Arc, RwLock},
+    thread::JoinHandle,
+};
+
+use super::dap_types;
+use super::SequenceNumber;
+
+enum StreamState {
+	// The client is waiting for a Stream to be sent from the thread
```
SpaceManiac, line 17: Mixed indentation

src/langserver/debugger/auxtools.rs
```diff
@@ -0,0 +1,432 @@
...
+            proc: ProcRef {
+                path: path.to_owned(),
+                override_id,
+            },
+            offset,
+        })?;
+
+        match self.read_response_or_disconnect()? {
+            Response::LineNumber { line } => Ok(line),
+            _ => Err(Box::new(super::GenericError("received incorrect response"))),
```
SpaceManiac, line 160: These error messages should contain at least what was expected and what they got instead

src/langserver/debugger/auxtools.rs
```diff
@@ -0,0 +1,432 @@
...
+                }
+                
+				self.run(stream);
+			}
+
+			Err(e) => {
+				eprintln!("Debug client failed to accept connection: {}", e);
+			}
+		})
+    }
```
SpaceManiac, line 343: mixed indentation getting worse

SpaceManiac (commented):
Is there any chance that the `/world/Del` hook can be stuffed into the .dll itself rather than seeping into DM? What about `stack_trace`?

> we can have multiple separate libraries loaded at once.

Can you explain how this works? Would I be shipping a "just the debugger" DLL and require users to install the base separately, or would I be shipping a base+debugger DLL where the base might or might not be the one "in charge"? What are the failure modes?

> I accidentally ran a `cargo fmt` on the files I was working with at some point, so there's a few formatting changes about.

I'll live, but please avoid this in the future. It does add noise to the diff.

src/langserver/Cargo.toml
SpaceManiac: I will probably revert this in the name of not having to modify my deploy scripts, but I suggest filing an up-to-date report upstream

src/langserver/debugger/auxtools.rs
willox: wouldn't be a problem if running `cargo fmt` didn't try to fix up every file in the repo ;_;

willox:
> Is there any chance that the /world/Del hook can be stuffed into the .dll itself rather than seeping into DM? What about stack_trace?

Doing the cleanup from a hooked proc is a little difficult because the code isn't really in a state where the hooks can be removed. Using the FFI interface was a simple way to get around that. It doesn't just disconnect the debug server, but also gets auxtools into a state where starting another DMB won't fuck up.

Auxtools modules do not unload/reload when restarting the world.

> Can you explain how this works? Would I be shipping a "just the debugger" DLL and require users to install the base separately, or would I be shipping a base+debugger DLL where the base might or might not be the one "in charge"? What are the failure modes?

The base is statically linked in. You can just ship the dll and everything will work as long as the SpacemanDMM.toml and DM have the necessary stuff. There's a small performance loss on proc call hooking for each library loaded, but except for when using the debug server I don't expect any code-bases to load more than one at a time.

I didn't mention above, but the `call(debug_server, "auxtools_init")()` will gracefully return `"FAILURE (<message here>)"` if it fails to load and `"SUCCESS"` otherwise.

src/langserver/debugger/mod.rs
willox: I'm not too sure on a good way to do this. It seems `struct Engine` does have a context (I'd noticed this but didn't realise it actually loaded the config).

Perhaps instead of passing the context to Debugger:new I can just pass the `engine` config's value.

src/langserver/debugger/auxtools.rs
willox: Output is now like:
`[main] Error responding to "threads": received unexpected response: expected Stacks, got Offset { offset: Some(112) }`

willox:
Fixed up the indentation you mentioned, there could easily be more.

src/langserver/Cargo.toml
willox: That's fine. It's more that I forgot to remove it first, but it was worth pointing out anyway.

Cyberboss:
Based PR

ZeWaka:
Based

willox:
> Is there any chance that the /world/Del hook can be stuffed into the .dll itself rather than seeping into DM? __What about stack_trace?__

I put a little thought into stack_trace, but we don't really have a way to trigger a runtime in BYOND without calling some sort of proc as auxtools has no way to manually create a new stack frame. If we just cause a runtime to happen at the end of our proc hook, it'll be treated as the runtime occurring in the caller and will cease execution of the proc that is calling us.

src/langserver/debugger/mod.rs
willox: Ok that's done. Hopefully it's fine.

SpaceManiac: Looks good.

SpaceManiac:
Still working through my review of auxtools itself.

> We can add [a DLL setting], but changing that option just seems more complicated than setting the env-var.

I'll take care of it myself if I find myself needing it.

> except for when using the debug server I don't expect any code-bases to load more than one at a time.

Well, I do have to worry about that situation. Is there a risk of problems if whichever of auxtools-tg.dll or debug_server.dll loads first has an older auxtools bundled? Maybe you can point me to the implementation in auxtools.

> debug server written in auxtools (currently located at https://github.com/willox/auxtools.)

Do you intend to maintain it going forward, or would it make sense to move it in-tree with SpacemanDMM eventually?

willox:
> Well, I do have to worry about that situation. Is there a risk of problems if whichever of auxtools-tg.dll or debug_server.dll loads first has an older auxtools bundled? Maybe you can point me to the implementation in auxtools.

There's no part of auxtools where two modules communicate with each other. We've just been writing it in a way where they won't interfere with each other to the point where it breaks. When a BYOND C-function is being hooked, we make sure to scan for a reference to it instead of for the function itself. That's basically it.

The side-effects should be pretty simple because of that. Stuff like two modules hooking a single proc will result in only the _last_ module to load ever seeing calls to that proc.

I've enforced a pretty simple rule in auxtools where we are only hooking BYOND functions that we absolutely have to. It's one of the older pieces of code, but if you check out `hooks.rs`'s `init` function you'll see we only hook `Runtime` and `CallProcById`.

The biggest risk of backwards incompatibility is if we have a signature in an old version of auxtools that is of a function's contents and a new version of auxtools hooks that function. It's something I intend to just avoid, though. Less hooks = good.

> Do you intend to maintain it going forward, or would it make sense to move it in-tree with SpacemanDMM eventually?

I'd expect to keep it working regardless of where it lives. It'd be easier for it to be in this repository, but at this point I'm more interested in getting auxtools on some servers than spending time moving the debug server.

SpaceManiac:
> When a BYOND C-function is being hooked, we make sure to scan for a reference to it instead of for the function itself.

Thanks for the explanation. Sounds like it will keep problems to a minimum and now I can at least start to investigate if something does go wrong.

> at this point I'm more interested in getting auxtools on some servers than spending time moving the debug server.

Sure, I was meaning more like after things have stabilized enough that /tg/station has migrated.

> Everything works out atm if the debug server loads first.

Will we be okay if the debug server loads second, but whatever loads first doesn't have a `#[runtime_handler] fn`?

src/langserver/debugger/evaluate.rs
```diff
@@ -1,45 +1,68 @@
...
 
-                let bytecode = extools.bytecode(&frame.proc, frame.override_id);
-                return Ok(EvaluateResponse::from(Self::format_disassembly(bytecode)));
+                        let bytecode = extools.bytecode(&frame.proc, frame.override_id);
+                        return Ok(EvaluateResponse::from(Self::format_disassembly(bytecode)));
+                    }
+                }
             }
+
+            DebugClient::Auxtools(_) => {}
```
SpaceManiac, line 39: Just a TODO or is there something making this difficult?

willox:
> Will we be okay if the debug server loads second, but whatever loads first doesn't have a #[runtime_handler] fn?

Nope. It won't crash or anything, but you will see extra runtimes coming from hooked proc calls when there's a runtime caught by our try/catch system.

The reasoning is that the runtime hook is defined as 
```cpp
extern "C" void runtime_hook(char* pError) {
	const char* pErrorCorrected = (pError != nullptr) ? pError : "<null>";
	if (runtime_contexts.top()) {
#ifdef USE_SJLJ
		longjmp(*current_jmp, 1);
#else
		throw AuxtoolsException(pErrorCorrected);
#endif
		return;
	}

	on_runtime(pErrorCorrected);
	return runtime_original(pError);
}
```

`on_runtime` is the part that calls into all the Rust runtime handlers. If your library is the second one to load, `runtime_original` will actually be pointing to the first library's hook. This means that your library will have called `on_runtime` before giving the first library a chance to catch and re-throw the runtime to itself.

src/langserver/debugger/evaluate.rs
willox: There's just no commands to add yet. You could add a diassemble command fairly easily, but I have only designed our disassembler to produce the correct length for each opcode/operands. The output is mostly correct, but that was a secondary objective.

The disassembler produces more correct output than extools, but it's not formatted. There's still some issues with GetVar (I've only noticed them in (init) methods so far.)

Example output would be:
```
Dism for Proc(/world/proc/enable_debugger)
	0-1: DbgFile(spaceman_dmm.dm)
	2-3: DbgLine(32)
	4-6: PushVal("env")
	7-9: PushVal("EXTOOLS_DLL")
	10-15: Call(RuntimeProcField(World, [], GetConfig), 2)
	16-18: SetVar(Local(0))
	19-20: DbgLine(33)
	21-23: GetVar(Local(0))
	24-24: Test
	25-26: Jz(38)
	27-28: DbgLine(34)
	29-31: GetVar(Local(0))
	32-34: PushVal("debug_initialize")
	35-36: CallName(ParamCount(0))
	37-37: Pop
	38-38: End
```

SpaceManiac: The formatting doesn't really matter, but the `#disassemble` command helped me a lot when working on extools.

It's not a priority for this PR, but I wanted to be sure there wasn't anything fundamental blocking it.

The DAP Disassemble request doesn't matter because VSC doesn't implement it.

willox: I'll add it in when I get around to figuring out what's wrong with (init) method SetVar instructions (it wasn't GetVar, but they have similar operands so the bug probably exists in both.)

SpaceManiac (approved):
Still some distance from production-ready but this PR is itself mergeable

PR #233, opened 2020-11-29
Auxtools update

willox:
This fulfils some auxtool debugger requests you had.
Depends on: https://github.com/willox/auxtools/commit/b3fb2f7f325cea6a5709540c5fa1687d02d574d4.

1) All connection modes except for `BACKGROUND` wait for the DAP client to be configured before continuing.
2) stddef.dm contents are sent to the debug client
3) disassemble eval command works (with the added benefit of being able to disassemble procs that aren't currently running)

The updated auxtools also does some other stuff you wanted:
1) src/usr moved to arguments
2) your PR https://github.com/willox/auxtools/pull/11
3) arguments with no formal parameter in a proc that is being debugged should show up

### DIFF
--- a/src/langserver/debugger/auxtools.rs
+++ b/src/langserver/debugger/auxtools.rs
@@ -165,6 +165,27 @@ impl Auxtools {
         }
     }
 
+    pub fn disassemble(&mut self, path: &str, override_id: u32) -> Result<String, Box<dyn std::error::Error>> {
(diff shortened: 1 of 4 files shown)


willox:
wait i forgot to do linux

SpaceManiac:
- it would be good if the `#dis` output included the original opcodes instead of DebugBreaks
- `#dis` output for OutputFormat instruction appears to print nonsense rather than the actual format string

PR #240, opened 2021-02-01
514 builtins support

ZeWaka:
Don't know if you want to put this based on a 514 branch or something, for full support. Here's the bulk of the work done for the builtins though.

- [x] gradient()
- [x] animate(space) // weird form
- [x] rgb(space) // weird form
- [x] rgb2num()
- [x] filter(type="bloom")
- [x] load_resource()
- [x] splicetext()
- [x] list.Splice()
- [x] time2text()
- [x] world.timezone
- [x] client.timezone
- [x] ||=
- [x] &&=
- [x] world.movement_mode + enums
- [x] := walrus/assign-into/expression assign
- [ ] ?[] // byond my abilities
- [x] world.map_cpu
- [x] particles/
- [x] generator/
- [x] embedded winget syntax // don't think we need to add anything
- [x] mouse actions // don't think we need to add anything

note: uncovered http://www.byond.com/forum/post/2652280

notes:

- didn't know for sure the right order for the PUNCT_TABLE, it says `shorter entries be first,` but it is unclear what it means by shorter - might mean sub-string length (if that's even the right word)?

Tested the produced executable:
image


src/dreammaker/lexer.rs
```diff
@@ -122,6 +124,7 @@ table! {
     "{\"", BlockString;
     "|",   BitOr;
     "|=",  BitOrAssign;
+    "||=",  SafeBitOrAssign;
```
SpaceManiac, line 127: The meaning of "requires that [...] shorter entries be first" is that `||` must be before `||=` in the list

SpaceManiac:
> ?[] // beyond my abilities

will think on this one but I suspect it will end up being a new `?[` punctuation named SafeLBracket or something like that

src/dreammaker/lexer.rs
```diff
@@ -77,6 +77,7 @@ table! {
     "%=",  ModAssign;
     "&",   BitAnd;
     "&&",  And;
+    "&&=", SafeBitAndAssign;
```
SpaceManiac, line 80: Wouldn't `AndAssign` be a more accurate name? If `x &&= y` is like `x = x && y`, then it's not "safe" and it's not "bit". Ditto `||=`

src/dreammaker/lexer.rs
```diff
@@ -99,6 +100,7 @@ table! {
     "//",  LineComment;
     "/=",  DivAssign;
     ":",   Colon -> CloseColon;
+    ":=",  ExpressAssign;
```
SpaceManiac, line 103: Seems like the release notes use the name "assign into" for this

src/dreammaker/lexer.rs
ZeWaka: Yeah. That would make a lot more sense. Was thinking of how ?[] works for some reason when I named these.

src/dreammaker/lexer.rs
```diff
@@ -291,6 +294,7 @@ impl Token {
                 QuestionMark |
                 BitXorAssign |
                 BitOrAssign |
+                SafeBitOrAssign |
```
SpaceManiac, line 297: Missing `&&=` and `:=` here, although this function is just for debugging/pretty-printing

SpaceManiac:
Thanks for the PR. Can punt on `?[` until later if you like

src/dreammaker/lexer.rs
ZeWaka: Yeah, but other languages (py) use "Assignment Expressions", so I figured that was the more language-agnostic terminology.
https://www.python.org/dev/peps/pep-0572/

SpaceManiac: `:=` being "assignment expression" is a Pythonism relating to the fact that in Python, `=` is a statement, whereas in BYOND `=` is already an expression

src/dreammaker/lexer.rs
ZeWaka: Yeah, I got confused about the ordering in the middle of adding this...

src/dreammaker/lexer.rs
ZeWaka: Hmm. So shall I change this to `IntoAssign`? - seems like the rest are reverse-styled as such.

SpaceManiac: Either word order is good enough, but I like `AssignInto` better. It's not Into/Assign like `+=` is Add/Assign, it's its own thing: Assign Into

src/dreammaker/builtins.rs
```diff
@@ -434,6 +453,7 @@ pub fn register_builtins(tree: &mut ObjectTree) {
         proc/get_step_rand(Ref);
         proc/get_step_to(Ref,Trg,Min=0);
         proc/get_step_towards(Ref,Trg);
+        proc/gradient(Gradient, index); // unsure how to handle (Item1, Item2, ..., index) form
```
ZeWaka, line 456: curious on if there's a proper way to address the two forms at once

src/dreammaker/builtins.rs
```diff
@@ -376,7 +394,7 @@ pub fn register_builtins(tree: &mut ObjectTree) {
             alpha, color, infra_luminosity, layer, maptext_width, maptext_height,
             maptext_x, maptext_y, luminosity, pixel_x, pixel_y, pixel_w, pixel_z,
             transform, dir, icon, icon_state, invisibility, maptext, suffix, appearance,
-            dir, radius,
+            dir, radius, space,
```
ZeWaka, line 397: note: if `space` is specified, `color` must be as well - potential linting/testing in the future?

src/dreammaker/builtins.rs
```diff
@@ -139,9 +140,9 @@ pub fn default_defines(defines: &mut DefineMap) {
         ELASTIC_EASING = Int(5);
         BACK_EASING = Int(6);
         QUAD_EASING = Int(7);
+        JUMP_EASING = Int(7); // 513
```
ZeWaka, line 143: note: fix once lummox fixes this (or resolve if he says it's not a bug)

src/dreammaker/builtins.rs
```diff
@@ -1060,6 +1088,35 @@ pub fn register_builtins(tree: &mut ObjectTree) {
...
         regex/proc/Find_char(text, start, end);
         regex/proc/Replace_char(text, rep, start, end);
+
+        // 514 stuff
+
+        generator;
+        generator/proc/Rand();
+        generator/proc/Turn(a);
+
+        particles;
```
ZeWaka, line 1098: these are annotated as `A datum (not derived from /atom)` which causes the ref lookup for this to be weird - something to investigate

...so it looks like there's also a var on movable atoms named the same thing 🤦‍♂️ 

also there's a reference error related: http://www.byond.com/forum/post/2652272

src/dreammaker/builtins.rs
```diff
@@ -1060,6 +1088,35 @@ pub fn register_builtins(tree: &mut ObjectTree) {
 
         regex/proc/Find_char(text, start, end);
         regex/proc/Replace_char(text, rep, start, end);
+
+        // 514 stuff
+
+        generator;
```
ZeWaka, line 1094: it'd be nice to have an enum for the various `type`s, and the applicable defines for rand being an enum as well

ZeWaka: not sure how to handle the operator overloads for this either

src/dreammaker/builtins.rs
```diff
@@ -1060,6 +1088,35 @@ pub fn register_builtins(tree: &mut ObjectTree) {
...
+        generator/proc/Rand();
+        generator/proc/Turn(a);
```
ZeWaka, line 1096: there's also a weird operator_turn in relation to this

src/dreammaker/builtins.rs
SpaceManiac: no, but it really only affects the popup parameter list completion, so it hasn't been a priority

src/dreammaker/builtins.rs
SpaceManiac: possible feature for the in-depth `filter()` verifier

src/dreammaker/builtins.rs
SpaceManiac: the existence of operator overloads for builtins isn't currently described by SpacemanDMM

ZeWaka:
no idea how we're going to support `matrix/proc/operator:=(m)`, `sound/proc/operator:=(sound/sound)`, and `icon/proc/operator:=(icon)`

PR #242, opened 2021-02-03
Optimize map rendering.

PJB3005:
It is much faster now, way less terrible blitting code.


SpaceManiac:
Reposting from Discord for future reference:
> PJB:
>> is it a problem that I changed the internal representation of `Image` to a `Vec<Rgba8>` which is a breaking change (and isn't a 2D array for easy handling)?
>
> SpaceManiac:
>> It's fine as long as the project compiles/passes tests/works. I make no API stability guarantees if used as a library
>> *and if the change is good

SpaceManiac:
Renders MetaStation about 9% faster for me which seems worthwhile

PR #252, opened 2021-03-10
Add detection of sleeping client procs

### LINKED ISSUE #236: client.SoundQuery() should trigger sleep lint
it has to wait for the client to reply

spookydonut:
MeasureText() and SoundQuery() technically sleep/block.

Fixes #236

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -1718,6 +1718,7 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
                     | "winget") {
                         self.env.sleeping_procs.insert_violator(self.proc_ref, unscoped_name, location);
                 }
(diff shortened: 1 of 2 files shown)


src/dreamchecker/lib.rs
```diff
@@ -1873,6 +1874,23 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
         }
     }
 
+    fn check_type_sleepers(&mut self, ty: TypeRef<'o>, location: Location, unscoped_name: &Ident) {
+        println!("{}, {}", ty, unscoped_name);
```
SpaceManiac, line 1878: Garbage on stdout will kill the langserver. You can use `eprintln!()` to avoid this problem (or delete this line since it's just debugging junk)

spookydonut: ooops

PR #254, opened 2021-03-12
Add 514 rgb() support

ZeWaka:
Allows for `rgb()` with arguments of form `[r,g,b|h,s,[v|l|y]],(a),(space)`

Also allows the full names of the arguments.
Technically, BYOND supports just the first letter so `halifax` works for `hue` but fuck that.

Also now lints to see if the values are in the valid ranges.

image
image
image
image


src/dreammaker/constants.rs
```diff
@@ -744,8 +744,8 @@ impl<'a> ConstantFolder<'a> {
                 "arccos" => self.trig_op(args, f32::acos)?,
                 "rgb" => {
                     use std::fmt::Write;
-                    if args.len() != 3 && args.len() != 4 {
-                        return Err(self.error(format!("malformed rgb() call, must have 3 or 4 arguments and instead has {}", args.len())));
+                    if args.len() != 3 && args.len() != 4 && args.len() != 5 {
```
SpaceManiac, line 747: Just tweaking the validation is insufficient here. The implementation needs to actually handle the HSV spaces as well.

src/dreammaker/tests/constants_tests.rs
```diff
@@ -1,6 +1,6 @@
 extern crate dreammaker as dm;
 
-use dm::constants::*;
+use dm::{constants::*};
```
ZeWaka, line 3:
```suggestion
use dm::constants::*;
```
webedit

src/dreammaker/constants.rs
```diff
@@ -744,19 +745,198 @@ impl<'a> ConstantFolder<'a> {
...
+                        }
+
+                        if let Some(i) = to_check.to_int() {
+                            if !range.contains(&i) {
+                                return Err(self.error(format!("malformed rgb() call, {} is not within the valid range ({}..{})", i, range.start(), range.end()))
+                                    .set_severity(Severity::Warning)
+                                    .with_location(self.location)
+                                );
+                            }
+                            let clamped= std::cmp::max(::std::cmp::min(i, *range.end()), *range.start());
```
ZeWaka, line 902:
```suggestion
                            let clamped = std::cmp::max(::std::cmp::min(i, *range.end()), *range.start());
```

SpaceManiac (approved):
Accepting this functionality as-is + some non-functional code cleanup. Looking in to extra tests for weirdo cases.

PR #255, opened 2021-03-15
panic! supports formatting

ZeWaka:
title, was a warning, rust2021 screamed at me


spookydonut:
I really need to rewrite these test helpers to be better

PR #257, opened 2021-03-23
MiniExpr no longer supports list indexing

willox:
I noticed that this was different from BYOND's behaviour where `new x[y]` actually compiles as `(new x)[y]`. The same behaviour applies for `new x?[y]` so I don't anticipate the code I am removing being needed for that.

I confirmed that BYOND has been compiling it this way since at least 513.1490, so if it changed it was long ago.


SpaceManiac (approved):
As far as I can tell, there was no BYOND change and this code was wrong all along.

PR #262, opened 2021-03-25
support for L?[x] list access syntax

willox:
I renamed the current `IndexKind` enum to `PropertyAccessKind` purely based off of the fact that somewhere in the byond ref both procs and vars are referred to as a datum's "properties" and I had no better name ideas.

The checker treats conditional list accesses exactly the same as normal ones. I was going to add some linting stuff for them but their expected behaviour is completely up in the air at the moment (see http://www.byond.com/forum/post/2667247.)


SpaceManiac:
The BYOND forum link in your comment is broken, do you have another reference?

SpaceManiac (approved):

willox:
It was a bug report showing how a bunch of stuff compiled incorrectly in the current stable/beta branches. Lummox is still changing `?.` in every single patch it seems so it's hard to decide on what is an error case atm.

It was merged into a similar thread but I can't find it.

PR #263, opened 2021-03-28
auxtools debug server: expression evaluation

willox:
The changes for the language server are minimal. This will require shipping version 2.0.0 of the debug server: ~https://github.com/willox/auxtools/releases/tag/debug-v2.0.0~ (use ~https://github.com/willox/auxtools/releases/tag/v2.1.0~ (use https://github.com/willox/auxtools/releases/tag/v2.2.0))

There are plenty of compiler features missing on the debug server's end, but it is definitely more than usable so I want to get it published. I plan on going a lot further this all of this, maybe we'll even have hot loading one day.

Noticeable differences from actual BYOND are:
1) It only compiles expressions! This means no for loops, no if statements, etc.
2) It lacks all static typing - the compiler has no access to type definitions and will treat all `.` accesses as `:` accesses and ignore the rules of const.
3) There are no implicit uses of `src`. This comes from the fact that there is no static typing, so to access a field on `src` you have to explicitly use `src.xyz` syntax.
4) Strings are lame. There is no interpolation support and the escape sequences are super limited too.
5) Some built-in procs aren't supported, but most are. They'll gracefully error if you try to use them.
6) No pre-processor.

For other limitations, just view the error list: https://github.com/willox/dmasm/blob/9ba328ef8c4e45c64072198e72c412cd9f685609/src/compiler.rs#L45-L76

tl;dr: relatively well-working debug console eval, watch eval, and conditional breakpoints

### DIFF
--- a/src/langserver/debugger/auxtools.rs
+++ b/src/langserver/debugger/auxtools.rs
@@ -217,9 +217,10 @@ impl Auxtools {
         }
     }
 
-    pub fn set_breakpoint(&mut self, instruction: &InstructionRef) -> Result<BreakpointSetResult, Box<dyn std::error::Error>> {
+    pub fn set_breakpoint(&mut self, instruction: InstructionRef, condition: Option<String>) -> Result<BreakpointSetResult, Box<dyn std::error::Error>> {
         self.send_or_disconnect(Request::BreakpointSet {
-            instruction: instruction.clone(),
+            instruction,
+            condition
         })?;
 
         match self.read_response_or_disconnect()? {
--- a/src/langserver/debugger/auxtools.rs
+++ b/src/langserver/debugger/auxtools.rs
@@ -165,10 +165,11 @@ impl Auxtools {
         }
     }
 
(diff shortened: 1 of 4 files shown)


willox:
I forgot to mention that this would require another proc to be added to code-bases.

```js
/proc/auxtools_expr_stub()
	return
```

The whole thing works by replacing this proc's bytecode at runtime and then calling it. If you are missing the proc, it just won't work. No crashes or anything.

willox:
I've released another debug server update since opening this PR, so I'd suggest shipping 2.1.0 instead of 2.0.0.

https://github.com/willox/auxtools/releases/tag/v2.1.0

willox:
another one: https://github.com/willox/auxtools/releases/tag/v2.2.0

SpaceManiac:
Looked good enough in testing to merge, but here's a few bug reports:
1. Seems like it's not possible to call verbs - they come out undefined instead
2. If eval hits a breakpoint, there's a None.unwrap() on `debug_server/src/server.rs:1203`
3. "runtime occurred when executing" message has an extra newline at the end

PR #264, opened 2021-04-23
Bumps up minor version to 1554.

AnturK:
Doesn't seem to be anything new since 49,  this is just to satisfy macros.


ZeWaka:
Wrong, world.profile was changed

AnturK:
Ah, good point. I'll have to update that.

SpaceManiac:
Happy to bump this if SpacemanDMM indeed fulfills that version, but

> just to satisfy macros

If you're referring to compile-time checks to prevent compiling with a broken version or anything like that, you should add `&& !defined(SPACEMAN_DMM)` to the condition.

PR #265, opened 2021-04-29
Lints for ambiguous `!` on the left hand side of a bitwise operation

ZeWaka:
title, essentially
Puts some helpful hints in order of likelihood. Could be intended to be: `!(x & y)` or `!x && y` or `~x & y`

Fulfils https://github.com/SpaceManiac/SpacemanDMM/projects/1#card-37357276

### DIFF
--- a/src/dreamchecker/lib.rs
+++ b/src/dreamchecker/lib.rs
@@ -1587,7 +1587,7 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
                 }
 
                 let rty = self.visit_expression(location, rhs, None, local_vars);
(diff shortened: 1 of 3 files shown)


src/dreamchecker/lib.rs
```diff
@@ -1642,6 +1642,12 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
             Expression::BinaryOp { op, lhs, rhs } => {
                 let lty = self.visit_expression(location, lhs, None, local_vars);
                 let rty = self.visit_expression(location, rhs, None, local_vars);
+                match op {
```
ZeWaka, line 1645: Could put this before the visitation above as well.

PR #266, opened 2021-05-15
Basic handling of relative type-paths to procs

Leshana:
Fixes false errors arising from relative paths to procs such as: _error: failed to resolve path .HasProximity_ even when HasProximity is defined on the current type.

Details:
- Adds a function that will attempt to resolve an identifer as a proc name if initial resolution fails; Returns NavigatePathResult so callers will know if it resolved to a proc or otherwise.
- Use that function when resolving '.', '/', or ':' path operators in navigate(), as those operators can resolve to a proc reference and modify navigate_path to handle the case when navigate() resolves to a proc refernce.
- Add unit tests covering common cases.
  - I drew test cases from the examples documented in SS13's _callback.dm_.

Example failure case:
```dm
/obj/machinery/camera/HasProximity(turf/T, atom/movable/AM, old_loc)
	// etc
/obj/machinery/camera/proc/upgradeMotion()
	sense_proximity(callback = .HasProximity)
```

### DIFF
--- a/src/dreamchecker/tests/static_type_tests.rs
+++ b/src/dreamchecker/tests/static_type_tests.rs
@@ -50,3 +50,44 @@ fn return_type() {
 "##.trim();
     check_errors_match(code, RETURN_TYPE_ERRORS);
 }
+
+pub const RELATIVE_PROC_PATHS_ERRORS: &[(u32, u16, &str)] = &[
+    (13, 17, "failed to resolve path .fake"),
+];
+
+// In particular we'll test the examples documented in SS13's callback.dm. None should error.
+#[test]
+fn relative_proc_paths() {
+    let code = r##"
+/proc/global_one()
+    return /proc/global_two         // absolute paths should always work
+
+/proc/global_two()
+    return .global_one              // global proc while in another global proc
+
+/mob/proc/test()
+    return
+
+/mob/proc/other()
+    var/proc_path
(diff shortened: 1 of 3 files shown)


SpaceManiac:
Does this correctly reject these cases?

```dm
/atom/proc/callee()
/atom/proc/caller()
/atom/movable/caller()
	world << .callee
```
```dm
/atom/proc/callee()
/atom/movable/proc/caller()
	world << .callee
```
```dm
/atom/movable/proc/callee()
/atom/proc/caller()
/atom/movable/caller()
	world << .callee
```

Also, is it worth encouraging `.ProcName` references when they might suddenly stop working if the peculiarities of which types override which procs (which otherwise doesn't matter) changes? I don't consider being bug-compatible with DM a goal of SpacemanDMM.

Leshana:
Hmm, actually those are *not* correctly found as errors.  In fact, it looks like detecting those would be quite difficult in the current SpacemanDMM architecture, as resolution depends on whether or not the lexically enclosing proc is a "proc" proc vs. an overridden proc.
That information doesn't seem to be easily available.

SpaceManiac:
Sounds like that's probably not worth the complexity. The question then is whether to err on the side of generous or strict. My first instinct is to go for strict - not accepting `.Foo` syntax even though DM will sometimes (according to esoteric rules) accept it.

PR #272, opened 2021-06-21
More tweaks to config stuff

AffectedArc07:
Note: I am still relatively new to rust, so chances are there is a lot of room for improvement here. Anyways.

This PR adds in a configuration for the `hide_invisible` render pass, to add extra types to it to forcibly hide. Configuration is as following
```toml
[map_renderer]
force_hide_types = [
    "/obj/effect/mapping_helpers",
    "/obj/effect/landmark",
    "/obj/effect/spawner/random_spawners",
]
```

This PR also adds a new render pass, `only-wires-and-pipes`. This is functionally a combination of `only-wires` and `only-pipes`, as you currently cannot stack both the passes.

### DIFF
--- a/src/tools/render_passes/mod.rs
+++ b/src/tools/render_passes/mod.rs
@@ -195,7 +195,7 @@ pub struct HideInvisible {
 
 impl RenderPass for HideInvisible {
     fn configure(&mut self, renderer_config: &dm::config::MapRenderer) {
-        self.overrides = renderer_config.force_hide_types.clone().into_iter().collect::<Vec<_>>();
+        self.overrides = renderer_config.hide_invisible.clone();
         // Put longer typepaths earlier in the list so that `/foo/bar` can override `/foo`.
         self.overrides.sort_unstable_by_key(|k| usize::MAX - k.len());
         // Append `/` to each typepath for faster starts_with later.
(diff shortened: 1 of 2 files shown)


src/tools/render_passes/mod.rs
```diff
@@ -188,13 +189,36 @@ impl RenderPass for HideAreas {
...
-pub struct HideInvisible;
+pub struct HideInvisible {
+    overrides: Vec<String>,
+}
+
 impl RenderPass for HideInvisible {
+    fn configure(&mut self, renderer_config: &dm::config::MapRenderer) {
+        self.overrides = renderer_config.force_hide_types.clone().into_iter().collect::<Vec<_>>();
+        // Put longer typepaths earlier in the list so that `/foo/bar` can override `/foo`.
+        self.overrides.sort_unstable_by_key(|k| usize::MAX - k.len());
```
AffectedArc07, line 200: Dont know if this is needed or not given that this list doesnt have overrides. Its just a path list.

src/tools/render_passes/mod.rs
```diff
@@ -391,6 +415,14 @@ impl RenderPass for Pipes {
     }
 }
 
+#[derive(Default)]
+pub struct WiresAndPipes;
+impl RenderPass for WiresAndPipes {
+    fn late_filter(&self, atom: &Atom, _: &ObjectTree) -> bool {
+        atom.istype("/obj/machinery/atmospherics/pipe/") || atom.istype("/obj/structure/cable/")
```
ZeWaka, line 422: Can this please be configurable?

PR #273, opened 2021-06-22
Bumped up minor version to 1556

Watermelon914:
https://github.com/tgstation/tgstation/pull/59771
minimum compile version is being changed to 1556 as a bunch of sendmap profiling seems to be utilised only on the version 1556 on tgstation, so these ifdef checks have been removed and the minimum compiler version has been brought up to 514.1556


Watermelon914:
nevermind this isn't needed, sendmaps profiling implemented in 1554

Watermelon914:
image

SpaceManiac:
while waiting on this to make it to a full release, /tg/ could also change the version requirement code to not run if `defined(SPACEMAN_DMM)`

PR #278, opened 2021-07-26
Better fonts

LetterN:
Uses a less "sharp" font. All the contents will be using Arial while codeblocks will use Consolas
image


Cyberboss (approved):
Bless

PR #285, opened 2021-09-17
Handle files with BOM in lexer.

PJB3005:
This came up in OpenDream (blame VS...). Figured it wouldn't be too hard to fix here.

Not sure if this is the best way to go about it, but it seems to work.

### DIFF
--- a/src/dreammaker/lexer.rs
+++ b/src/dreammaker/lexer.rs
@@ -483,7 +483,7 @@ enum Directive {
 }
 
 fn has_bom(slice: &[u8]) -> bool {
-    slice.len() > 3 && slice[0] == 0xEF && slice[1] == 0xBB && slice[2] == 0xBF
+    slice.starts_with(b"\xEF\xBB\xBF")
 }
 
 fn buffer_read<R: Read>(file: FileId, mut read: R) -> Result<Vec<u8>, DMError> {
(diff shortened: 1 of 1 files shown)


Cyberboss (approved):

src/dreammaker/lexer.rs
```diff
@@ -482,6 +482,10 @@ enum Directive {
     Stringy,
 }
 
+fn has_bom(slice: &[u8]) -> bool {
+    slice.len() > 3 && slice[0] == 0xEF && slice[1] == 0xBB && slice[2] == 0xBF
```
SpaceManiac, line 486: nit, but could use `.starts_with(b"\xEF\xBB\xBF")`

src/dreammaker/lexer.rs
```diff
@@ -630,9 +634,14 @@ impl<'ctx> HasLocation for Lexer<'ctx> {
 impl<'ctx> Lexer<'ctx> {
     /// Create a new lexer from a byte stream.
     pub fn new<I: Into<Cow<'ctx, [u8]>>>(context: &'ctx Context, file_number: FileId, input: I) -> Self {
+        let mut cow = input.into();
+        if has_bom(&cow) {
+            cow = Cow::from(cow[3..].to_owned());
```
SpaceManiac, line 639: Should be possible to subslice a Borrowed or to erase the first three bytes of an Owned, rather than always reallocating

SpaceManiac (changes requested):
BOMs are terrible and I dislike that VS still uses them

src/dreammaker/lexer.rs
PJB3005: What do you mean with erase? Set the BOM to null bytes? That would mess with unless I make the rest of the lexer ignore null bytes.

SpaceManiac: definitely a micro-optimization but
```rust
Cow::Owned(mut o) => { o.drain(..3); Cow::Owned(o) }
```

PJB3005: Ah I didn't realize that the `Owned` of a `Cow<[T]>` is a `Vec`. That makes sense yeah.

PR #293, opened 2021-10-29
use indexmap instead of linked hashmap, switch hasher to use ahash

jupyterkat:
ahash is straight up faster for string hashing
indexmap is at least maintained and has methods that we use, also faster iteration.

before:
before
after:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,9 +77,9 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
-version = "1.2.1"
+version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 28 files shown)


SpaceManiac:
I saw a much smaller speedup on my machine, over several trials: dreamchecker average time from 5.946s to 5.794s, dm-langserver average time from 6.488s to 6.344s. Not nothing, but not as drastic as you reported. Can you try doing a few extra trials so we can find out if it your numbers were a fluke (disk cache can be a big confounder), or if there's a difference between our machines?

jupyterkat:
might be because your machine just doesn’t support AES-NI like most modern machines do. have you tried on larger codebases where the extra seconds matter more

SpaceManiac:
As far as I can tell my processor does support AES-NI. I was testing on /tg/station.

PR #297, opened 2021-11-08
Fix auxtools cfg misspelling

igorsaux:
Found this misspelling which leads to not bundling auxtools lib at all.


SpaceManiac:
Ah, I probably didn't notice this because I usually build with either none or both

PR #302, opened 2022-02-07
Linter checks for the switch(rand(L, H)) pattern

pali6:
A pattern such as
```
switch(rand(1, 3))
  if(1)
    foo()
  if(2)
    bar()
  if(3)
    baz()
```
is not uncommon in SS13 code. Sadly, it is also not uncommon for people to add new cases to a longer switch of this form and forget to change the `rand` arguments at the top. This PR adds a linter warning for when the `rand` range is not fully covered by the cases and for when a case lies completely outside of the `rand` range.

This currently generates 7 warnings on tgstation master. Sadly, it is (at least to me) not quite clear if the omissions of a part of the `rand` range are intentional there or not. However, I believe that even if they are intentional it is worth adding the missing part of the range as a case with empty body explicitly to make the intent clear to future developers.

As a less related (but perhaps more important) addition this PR now also adds a warning for switch branches of the form `if(X || Y)` which is pretty much always a mistake and should instead be `if(X, Y)`. This lint generates 5 warnings on current tg code, seemingly all being actual mistakes.

### DIFF
--- a/crates/dreamchecker/src/lib.rs
+++ b/crates/dreamchecker/src/lib.rs
@@ -1484,7 +1484,7 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
                 for (case, ref block) in cases.iter() {
                     let mut scoped_locals = local_vars.clone();
                     if let [dm::ast::Case::Exact(Expression::BinaryOp{op: BinaryOp::Or, ..})] = case.elem[..] {
-                        error(case.location, format!("Elements in a switch-case branch separated by ||, this is likely in error and should be replaced by a comma"))
+                        error(case.location, "Elements in a switch-case branch separated by ||, this is likely in error and should be replaced by a comma")
                             .set_severity(Severity::Warning)
                             .register(self.context);
                     }
--- a/crates/dreamchecker/src/lib.rs
+++ b/crates/dreamchecker/src/lib.rs
@@ -1483,6 +1483,11 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
                 self.visit_expression(location, input, None, local_vars);
                 for (case, ref block) in cases.iter() {
                     let mut scoped_locals = local_vars.clone();
+                    if let [dm::ast::Case::Exact(Expression::BinaryOp{op: BinaryOp::Or, ..})] = case.elem[..] {
+                        error(case.location, format!("Elements in a switch-case branch separated by ||, this is likely in error and should be replaced by a comma"))
+                            .set_severity(Severity::Warning)
+                            .register(self.context);
+                    }
                     for case_part in case.elem.iter() {
                         match case_part {
                             dm::ast::Case::Exact(expr) => { self.visit_expression(case.location, expr, None, &mut scoped_locals); },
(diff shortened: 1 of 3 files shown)


pali6:
Warnings emitted on tgstation master for reference:
```
code/modules/mob/mob_helpers.dm, line 103, column 3:
warning: Switch branches on rand() with range 1 to 20 but no case branch triggers for 2

code/modules/mob/mob_helpers.dm, line 142, column 3:
warning: Switch branches on rand() with range 1 to 15 but no case branch triggers for 6

code/datums/brain_damage/severe.dm, line 182, column 2:
warning: Switch branches on rand() with range 1 to 6 but no case branch triggers for 6

code/modules/surgery/advanced/lobotomy.dm, line 54, column 2:
warning: Switch branches on rand() with range 1 to 4 but no case branch triggers for 4

code/modules/hydroponics/hydroponics_chemreact.dm, line 15, column 2:
warning: Switch branches on rand() with range 0 to 100 but no case branch triggers for 0

code/modules/vending/_vending.dm, line 500, column 4:
warning: Switch branches on rand() with range 1 to 100 but no case branch triggers for 26

code/modules/ruins/lavalandruin_code/elephantgraveyard.dm, line 137, column 2:
warning: Switch branches on rand() with range 1 to 8 but no case branch triggers for 8
```

crates/dreamchecker/src/lib.rs
```diff
@@ -1475,11 +1477,17 @@ impl<'o, 's> AnalyzeProc<'o, 's> {
...
             },
             Statement::Switch { input, cases, default } => {
+                check_switch_rand_range(input, cases, default, location, self.context);
                 let mut allterm = ControlFlow::alltrue();
                 self.visit_control_condition(location, input);
                 self.visit_expression(location, input, None, local_vars);
                 for (case, ref block) in cases.iter() {
                     let mut scoped_locals = local_vars.clone();
+                    if let [dm::ast::Case::Exact(Expression::BinaryOp{op: BinaryOp::Or, ..})] = case.elem[..] {
+                        error(case.location, format!("Elements in a switch-case branch separated by ||, this is likely in error and should be replaced by a comma"))
```
willox, line 1487: no need for format! here

SpaceManiac (approved):
Extremely good, thank you

KIBORG04:
Can you make it possible to turn off `warning Switch branches to rand() with a range from 1 to 6, but without branch triggers for 6` in configuration ? It is not always possible to solve this by inserting the missing IFs. Example:
```
/proc/some_func()
    for(var/i in 1 to 10)
        switch(rand(1, 20)
            if(1 to 2)
                ..code..
            if(3 to 15)
                ..code..
            if(16 to 20)
                ???
        do_post_switch()
```
If I replace `???` to `break` than get the wrong result, to `continiue` the same, to `return` the same. And there are no other options.

SpaceManiac:
I recommend leaving it blank with an explanatory comment:
```dm
switch(rand(1, 20))
    if(1 to 9)
        bad_effect()
    if(10)
        // Neutral roll, do nothing.
    if(11 to 20)
        good_effect()
```

This also works fine with `else` in the final position, but for some reason not on the final `if`, but you can use a lone semicolon in that case:
```dm
switch(rand(1, 20))
    if(1 to 17)
        get_hurt()
    if(18 to 20)
        ; // Defense succeeded, no damage.
```

PR #307, opened 2022-03-04
Lint initial(thing.var), where var is a list

### LINKED ISSUE #284: [Feature Request] Lint again initial(thing.var), where var is a list
The following code is never correct:

```dm
/datum/species
    var/list/allowed_traits = list(1, 2, 3)

/proc/do_something(datum/species/species_type)
    if (1 in initial(species.allowed_traits))
```

...because lists are made on `init`, the `initial` value is always null.

spookydonut:
Closes #284

### DIFF
--- a/crates/dreamchecker/src/lib.rs
+++ b/crates/dreamchecker/src/lib.rs
@@ -692,7 +692,7 @@ impl<'o> AnalyzeObjectTree<'o> {
                             .with_blocking_builtins(self.sleeping_procs.get_violators(*child_violator).unwrap())
                             .register(self.context)
                     }
-                } 
+                }
                 if let Some(calledvec) = self.call_tree.get(&nextproc) {
                     for (proccalled, location, new_context) in calledvec.iter() {
                         let mut newstack = callstack.clone();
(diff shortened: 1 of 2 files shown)


SpaceManiac (changes requested):
1. Error message should be more descriptive about what the issue is. Along the lines of:
    `initial() called on var/list/{}, but it always returns null for lists`
2. The `for` loop here will give redundant errors if you write `initial(a.b.c.d)`. Should probably just pull `.last()`.
3. It might make sense to apply this to anything that isn't a primitve, rather than lists, because `new /obj` likewise returns `null` from `initial()`.
4. Could go either way on this but it might make sense to analyze the actual initial values of that variable for subtypes to see if they are non-primitives, rather than just the static types.

SpaceManiac:
Checked up on this. I think checking the `static_ty` is the wrong approach. /tg/station has 97 detections, all of which are false positives, for example `var/datum/id_trim/trim`, then `trim = /datum/id_trim/away`, then `initial(trim)`.

I think instead we'd need to collect all vars referred to in `initial()` calls then see if any overrides of those vars are `Constant::New`, `List`, or `Call`.

Issue #267, opened 2021-05-21
Unrelated sleeping overrides hijack each other into making everything pass SHOULD_NOT_SLEEP (feat waitfor)


fira:
Vaguely similar to #247 and likely linked, but specific enough that it really looks like an implementation bug

Consider the following file:

```
/datum/proc/do_things()
        set SpacemanDMM_should_not_sleep = 1

/datum/proc/thing()
        do_complicated_thing()

/datum/sub/do_things()
        thing()

/proc/do_complicated_thing()
        usr.client.SoundQuery()
```

This works as expected:

```
onefile.dm, line 8, column 21:
error: /datum/sub/proc/do_things sets SpacemanDMM_should_not_sleep but calls blocking proc /proc/do_complicated_thing
- 3:2: SpacemanDMM_should_not_sleep set here
- 9:2: /datum/proc/thing() called here
- 6:2: /proc/do_complicated_thing() called here
- 12:12: client.SoundQuery() called here
```

Now strap at the end funny hijack proc, **on a different subtype**:

```
/datum/zzz/thing()
        sleep(10)
```

Result: 

```
Procs analyzed: 5. Errored: 0. Builtins: 335.

============================================================
Analyzing proc override validity...

============================================================
Analyzing proc call tree...

============================================================
Found 0 diagnostics
```

Funny huh? In our real world example, this was even more awkward from the fact this still generated a warning, that was also ignored by the same override setting waitfor=0

fira:
after testing looked to me like an oversight of visiting in `check_proc_call_tree` - the different checking blocks probably shouldn't be exclusive. otherwise you can get into a situation where there are sleeping overrides, so rest of call graph is not checked, but because all the sleeping overrides were actually relevant to a different sub-type this still isn't a fatal error and silently ignored. something similar also happens with waitfor as mentioned above, and just the general assumption regarding (intended or not but effectively) stopping on error.

PR #311, opened 2022-03-26
Fix Sleep/Purity checks being skipped under certain overrides

fira:
Long overdue PR for #267 and as discuted on coderbus ages ago back then

TL;DR is, there being an override one way or another shouldn't prevent you from checking children too

Discovered on CM-SS13 as we had a controller held up by sleep randomly in production, and no amount of hammering SHOULD_NOT_SLEEP checks seemed to trip anything !

As additional and separate note, and as we discussed back then on coderbus, there are more situations with assumptions of types that mean SpacemanDMM will miss out on a big part of the violators. It's actually for best, because "properly" checking is near impossible on SS13 due to double dispatch constructs just about erasing the "actual" types. This resulted in about 25,000 violating procs when fixing it for CM-SS13, for example...

### DIFF
--- a/crates/dreamchecker/src/lib.rs
+++ b/crates/dreamchecker/src/lib.rs
@@ -695,7 +695,8 @@ impl<'o> AnalyzeObjectTree<'o> {
                             .with_blocking_builtins(self.sleeping_procs.get_violators(*child_violator).unwrap())
                             .register(self.context)
                     }
-                } else if let Some(calledvec) = self.call_tree.get(&nextproc) {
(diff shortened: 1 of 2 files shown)


spookydonut:
@fira I'd like to see a test case written for this that proves it works and will fail if your changes are removed.

fira:
> @fira I'd like to see a test case written for this that proves it works and will fail if your changes are removed.

there is a snippet in the issue but let me add a test proper then too

PR #313, opened 2022-04-22
Use parking_lot to make dreammaker crate 100% thread safe

tigercat2000:
Needed for a downstream project that wants to throw dm::Context around different threads like a psycho.

All tests still pass, but I'd definitely suggest double checking my work here-- I'm still pretty new to Rust's threading system, and I'm not 100% sure that these are actually safe changes that will Just Work:tm:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,9 +56,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


SpaceManiac:
Why do you need to pass the `Context` itself around and not just the `ObjectTree` that parsing outputs?

tigercat2000:
Convenient to access the file list, maybe? I'm pretty sure the downstream figured out how to pass the subset they actually need between threads anyways...

tigercat2000:
~~I guess this really dooooesn't need the potential performance impact of Arc~~

PR #315, opened 2022-04-30
Fix Clippy lints

tigercat2000:
This resolves every single error raised by [Clippy](https://github.com/rust-lang/rust-clippy) for every active crate. See commits for details, it's not really possible to list the ~240 odd issues Clippy saw- most are formatting or redundant code, there's only a few things that are functionally better.

### DIFF
--- a/crates/dm-langserver/src/debugger/extools.rs
+++ b/crates/dm-langserver/src/debugger/extools.rs
@@ -126,6 +126,8 @@ impl ExtoolsHolder {
     }
 
     pub fn disconnect(&mut self) {
+        // This part of code is not complete, we don't want to use matches!
+        #[allow(clippy::single_match)]
         match std::mem::replace(&mut self.0, ExtoolsHolderInner::None) {
             ExtoolsHolderInner::Attaching { cancel_tx, .. } => { let _ = cancel_tx.send(()); },
             // TODO: ExtoolsHolderInner::Listening
--- a/crates/dm-langserver/src/debugger/extools.rs
+++ b/crates/dm-langserver/src/debugger/extools.rs
@@ -338,8 +338,8 @@ impl Drop for Extools {
 
 fn parse_lineno(comment: &str) -> Option<i64> {
     let prefix = "Line number: ";
-    if comment.starts_with(prefix) {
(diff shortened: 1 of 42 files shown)


crates/dm-langserver/src/debugger/mod.rs
```diff
@@ -1089,7 +1093,7 @@ handle_request! {
             DebugClient::Extools(extools) => {
                 let extools = extools.get()?;
 
-                if params.variablesReference >= 0x01_000000 {
+                if params.variablesReference >= 0x0100_0000 {
```
willox, line 1096: I'm not particularly fussed here but the strange _ from before would likely have been separating the byond type id from the object identifier.

1 byte for the type, 3 bytes for the rest

tigercat2000: i was wondering if that was the case but the fact that there's no comments anywhere in sight indicating that was a pretty big 🤷🏻

SpaceManiac:
I should probably fix this separately but I noticed the Hash/Eq impls for Prefab and Pop only take the keys into account and not the values.

crates/dmdoc/src/main.rs
```diff
@@ -919,10 +919,7 @@ fn module_entry<'a, 'b>(modules: &'a mut BTreeMap<String, Module1<'b>>, path: &P
     modules.entry(module_path(path)).or_insert_with(|| Module1 {
         htmlname: module_path(path),
         orig_filename: path.display().to_string().replace('\\', "/"),
-        name: Default::default(),
-        teaser: Default::default(),
-        items_wip: Default::default(),
-        defines: Default::default(),
+        .. Default::default()
```
tigercat2000, line 922: Woah, I didn't know you could do this, this is neat.

crates/dm-langserver/src/debugger/extools.rs
```diff
@@ -339,9 +339,8 @@ impl Drop for Extools {
 }
 
 fn parse_lineno(comment: &str) -> Option<i64> {
-    let prefix = "Line number: ";
```
tigercat2000, line 340: lmao I guess I wasn't paying much attention here

tigercat2000:
👍🏻

PR #316, opened 2022-05-07
Add [u8] read/write functions to Image/IconFile/Metadata

tigercat2000:
This is so IconDiffBot2 can operate purely in memory.

Metadata::from_raw and IconFile::from_raw take an AsRef<[u8]>
as the sequence of bytes of a valid PNG+DMI file to produce their types.

Image::to_raw gives a Vec<u8> which is just the output from
png::Encoder.

There is no Image::from_raw because there's no reason not to use
IconFile::from_raw, and there is no to_raw for Metadata or IconFile
because they don't support to_file either. IconFile could, but it's
unnecessary.

### DIFF
--- a/crates/dmm-tools/src/dmi.rs
+++ b/crates/dmm-tools/src/dmi.rs
@@ -30,11 +30,7 @@ pub struct IconFile {
 
 impl IconFile {
     pub fn from_file(path: &Path) -> io::Result<IconFile> {
-        let (bitmap, metadata) = Metadata::from_file(path)?;
-        Ok(IconFile {
-            metadata,
-            image: Image::from_rgba(bitmap),
-        })
+        Self::from_raw(std::fs::read(path)?)
     }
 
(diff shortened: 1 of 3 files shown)


crates/dmm-tools/src/dmi.rs
```diff
@@ -159,6 +167,21 @@ impl Image {
         Ok(())
     }
 
+    #[cfg(feature = "png")]
+    pub fn to_raw(&self) -> io::Result<Vec<u8>> {
+        let mut vector = Vec::new();
+        {
+            let mut encoder = png::Encoder::new(&mut vector, self.width, self.height);
```
tigercat2000, line 174: is this safe? It produces an `Encoder<&mut Vec<u8, Global>>`, but I'm wary about the `Write` trait and references to Vec's since there are some circumstances where it'll just blindly write to a vector without updating the size

tigercat2000: apparently it's fine

a

SpaceManiac:
I'd suggest putting the common code in generic from_read and to_write methods rather than copy-pasting.

PR #318, opened 2022-05-12
IconFile -> Gif/Png Renderer

tigercat2000:
Needs to be rebased onto master, I'm still working on making the API perfect

### DIFF
--- a/crates/dmm-tools/src/dmi.rs
+++ b/crates/dmm-tools/src/dmi.rs
@@ -124,11 +124,11 @@ impl IconFile {
                     // aka centiseconds. We're measuring in BYOND ticks, aka deciseconds.
                     // 1 decisecond = 10 centisecond, so we multiply by 10.
                     // And it's a u16 for some reason so we just SHRUG and floor it.
-                    frame.delay = (if let Some(vec) = &render.delays {
-                        *vec.get(index).unwrap_or(&1.0)
-                    } else {
-                        1.0
-                    } * 10.0) as u16;
+                    frame.delay = (render
+                        .delays
+                        .as_ref()
+                        .map_or_else(|| 1.0, |f| *f.get(index).unwrap_or(&1.0))
+                        * 10.0) as u16;
                     // the disposal method by default is "keep the previous frame under the alpha mask"
                     // wtf
                     frame.dispose = DisposalMethod::Background;
--- a/crates/dmm-tools/src/dmi.rs
+++ b/crates/dmm-tools/src/dmi.rs
@@ -20,6 +20,7 @@ pub type Rect = (u32, u32, u32, u32);
 // ----------------------------------------------------------------------------
 // Icon file and metadata handling
 
+#[derive(Clone)]
 pub struct RenderResult {
     pub frames: Vec<Image>,
     pub delays: Option<Vec<f32>>,
--- a/crates/dmm-tools/src/dmi.rs
+++ b/crates/dmm-tools/src/dmi.rs
@@ -2,7 +2,7 @@
 //!
 //! Includes re-exports from `dreammaker::dmi`.
 
-use std::{io, iter::FromIterator};
+use std::{io};
 use std::path::Path;
 use bytemuck::Pod;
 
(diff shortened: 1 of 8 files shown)


tigercat2000:
This is finally something I'm comfortable with having added upstream, the API is stabilized and functions well for IDB2's use-case, but should be able to be used by other projects as well.

New crates:
 - Added the `either` crate to get around an annoying type issue with chained iterators.
 - Added the `derivative` crate to make `State` realistically comparable across different icon files (important for diffing in IDB2)

Notes on public API changes:
 - Most of the new stuff is in the inline module `dmm_tools::dmi::render`.
 - `IconFile`, `Rgba8`, and `Image` are now `Debug`
 - `from_bytes` on `Image`, `IconFile`, and `State` no longer takes an explicit &[u8] array, instead it takes the trait bound `<B: AsRef<[u8]>>` for ease of "eat my ownership because I don't care"
 - `IconFile` has a new function `pub fn get_icon_state<S: AsRef<str>>(&self, icon_state: S) -> io::Result<&State>` for looking up states by name.
 - `Image` has a `clear()` method which fills the data array with defaults.
 - `State` has a new member, `pub duplicate: Option<u32>`, which is used for correctly supporting duplicate icon states by enumerating how many times the state name has been seen before this particular `State` was read.
 - `State` has a `get_icon_state` method which backs `IconFile`'s `get_icon_state`.
 - `State` has a `is_animated(&self) -> bool` method which is used in IDB2 for precalculating the file extension
 - `State` also has `get_state_name_index`, which is used by `parse_metadata` to give duplicate states a unique entry into the `state_names` `BTreeMap`.
 - `parse_metadata` now keeps track of duplicate states and puts them in `state_names` according to the above.

crates/dmm-tools/src/dmi.rs
```diff
@@ -60,6 +60,7 @@ pub mod render {
         /// This is here so that duplicate icon states can be handled by not relying on the btreemap 
         /// of state names in [`Metadata`]. 
         pub fn render_state<W: std::io::Write>(&self, icon_state: &State, target: W) -> io::Result<RenderType> {
+            // Note: we don't use [`State::is_animated()`] here because we actually need to destructure the tuples.
```
tigercat2000, line 63: I'm not in love with the fact that these two functions can experience divergence, since that'll break one of IDB's assumptions, but I have no idea how else to do this given the requirements to destructure here.

tigercat2000: As a countermeasure against broken images in my own code resulting from this possibility, I made it so that IDB errors when this happens. (Since `render_state` returns what it actually used, can compare "what actually happened" to "predicted file extension".)

tigercat2000: RIP "stable API", the correct solution to this, which I implemented in 3b610a8 and 23dbc72 is a guard pattern to allow the end user to do some things between prepping to render and rendering.

SpaceManiac:
non-thorough review items:
* Why an inline `mod`? It's nearly half the whole file now, seems like it could be its own `.rs`
* I'd prefer accepting just `&[u8]` or `&str` rather than `AsRef` generics because it reduces code size a bit and is simpler overall. AsRef makes more sense if you're writing a generic container or intend various reference-to-reference conversions

tigercat2000:
> non-thorough review items:
> 
>     * Why an inline `mod`? It's nearly half the whole file now, seems like it could be its own `.rs`

It should be, yeah, I started it when it was thin and didn't think about it. I'll fix this soon.

>     * I'd prefer accepting just `&[u8]` or `&str` rather than `AsRef` generics because it reduces code size a bit and is simpler overall. AsRef makes more sense if you're writing a generic container or intend various reference-to-reference conversions

This is just for ergonomics, I'd prefer to keep them- it allows passing off ownership of temporaries to these functions and letting them handle the drops as soon as possible instead of worrying about explicit references in IDB's code. 

>     * Are duplicate states important to IDB? I've been thinking of them more like an error condition

Yes, they need to be marked and rendered, both for compatibility with IDB1 and simply because it's useful to know in a diff if there are duplicate states, so that you can reject if they are just bloat/etc. The whole point is "not needing to look at the dmi file by hand."

SpaceManiac:
It's easy to do drops ASAP by writing something like `accepts_str_ref(&{owned_string});`.

The way you have state_names right now is going to cause issues if you have a .dmi with states named `foo`, then `foo1`, then `foo` again. I think it would make more sense to key off a `(String, u32)` tuple or something like that.

tigercat2000:
> &{owned_string}

Whoa, it took me a second to understand what this was doing, that's a really cool way to do it
Agreed on the state names

SpaceManiac:
Any opinions on dropping the "(0)" which is most of the time not useful?
- When dup = 0, just show the state name
- When dup > 0, show `{state_name} (duplicate #{dup})`

I may get bored and fiddle with the API on top of this but this looks mergeable.

tigercat2000:
@SpaceManiac done

PR #327, opened 2022-06-17
Check for empty docs_in in flush_docs

Mothblocks:
Before

Tracy_2022-06-16T17-17-28

After

Tracy_2022-06-16T17-18-09


Mothblocks:
I web-edited this because my actual code is full of tracey profiles and cargo fmt stuff

Mothblocks:
You're right, this is a typo.

PR #330, opened 2022-07-10
Requested feature, detect for loops over turf contents

spookydonut:


crates/dreammaker/src/config.rs
```diff
@@ -120,6 +120,10 @@ impl Config {
         Ok(toml::from_str(&config_toml)?)
     }
 
+    pub fn read_toml_string(toml: String) -> Result<Config, Error> {
+        Ok(toml::from_str(&toml)?)
```
SpaceManiac, line 124: If this takes `&str`, presumably so could the chain that calls it, rather than requiring `config.to_string()` in the tests

Issue #334, opened 2022-08-23
SHOULD_NOT_SLEEP and SHOULD_BE_PURE have false positives based on incorrect inheritance assumptions


out-of-phaze:
```
/proc/foo()
    SHOULD_NOT_SLEEP(TRUE)

/datum/proc/foo()
    sleep(5)
```
`/datum/proc/foo()` will report that it has `SHOULD_NOT_SLEEP` set when it shouldn't.

out-of-phaze:
![image](https://cdn.discordapp.com/attachments/484170915253321734/1011482943224557759/unknown.png)
An example in a real environment, although the proc didn't originally have the annotation--it was added for demonstration.

PR #335, opened 2022-08-23
Add diagnostics for mismatched argument type and count

out-of-phaze:
I don't know what I'm doiiiing!

A hacky first attempt at adding two diagnostics:
- `override_changed_type`: Raised for procs whose non-keyword arguments don't match (subtype/supertype) the argument with the same index on the parent.
- `override_fewer_arguments`: Raised for procs which have fewer non-keyword arguments than their parent. This is mainly useful for ensuring the langserver passes the arguments to later types, but can sometimes reveal issues linked with parent calls that manually specify arguments. (If I could have it report *those,* that'd be a dream.)

### DIFF
--- a/crates/dreamchecker/src/lib.rs
+++ b/crates/dreamchecker/src/lib.rs
@@ -354,11 +354,13 @@ fn run_inner(context: &Context, objtree: &ObjectTree, cli: bool) {
     objtree.root().recurse(&mut |ty| {
         for proc in ty.iter_self_procs() {
             analyzer.check_kwargs(proc);
+            analyzer.check_arg_type(proc);//count(proc);
             analyzer.propagate_violations(proc);
         }
     });
 
     analyzer.finish_check_kwargs();
+    analyzer.finish_check_arg_count();
 
     cli_println!("============================================================");
     cli_println!("Analyzing proc call tree...\n");
@@ -564,6 +566,8 @@ pub struct AnalyzeObjectTree<'o> {
 
     sleeping_overrides: ViolatingOverrides<'o>,
     impure_overrides: ViolatingOverrides<'o>,
+    decreasing_argc_overrides: ViolatingOverrides<'o>,
+    incorrect_arg_type_overrides: ViolatingOverrides<'o>,
 }
 
 impl<'o> AnalyzeObjectTree<'o> {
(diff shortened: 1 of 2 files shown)


crates/dreamchecker/src/lib.rs
```diff
@@ -354,11 +354,13 @@ fn run_inner(context: &Context, objtree: &ObjectTree, cli: bool) {
     objtree.root().recurse(&mut |ty| {
         for proc in ty.iter_self_procs() {
             analyzer.check_kwargs(proc);
+            analyzer.check_arg_type(proc);//count(proc);
```
ZeWaka, line 357:
```suggestion
            analyzer.check_arg_type(proc);
```
;)

crates/dreamchecker/src/lib.rs
```diff
@@ -845,6 +851,109 @@ impl<'o> AnalyzeObjectTree<'o> {
...
 
+    /// Check that a ProcRef doesn't have fewer arguments than its parent
+    pub fn check_arg_count(&mut self, proc: ProcRef<'o>) {
+        if proc.is_builtin()
+        {
+            return;
+        }
+        if let Some(parent) = proc.parent_proc()
+        {
+            if parent.is_varargs() /* remove this later, it's a hack for objtree proc parents including root */ || parent.ty().get().is_root() /* seriously remove it */
```
ZeWaka, line 862: line length
```suggestion
			/* remove this later, it's a hack for objtree proc parents including root */
            if parent.is_varargs() || parent.ty().get().is_root() 
            /* seriously remove it */
```

out-of-phaze: Only the is_root needs to be removed when #334 is solved. The varargs check is fine.

crates/dreamchecker/src/lib.rs
out-of-phaze: whoops. yeah I can probably remove the check_arg_count method entirely since I replaced it with check_arg_type, and I can rename it to check_args

crates/dreamchecker/src/lib.rs
ZeWaka:
```suggestion
            if parent.is_varargs() || 
            /* remove this later, it's a hack for objtree proc parents including root */
            parent.ty().get().is_root()
            /* seriously remove it */
```
this then, or just say that - it's wrong to have incredibly long lines with functionality hidden

PR #336, opened 2022-08-26
Bump up all dependencies

jupyterkat:
no other changes of note

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,12 +26,33 @@ dependencies = [
 ]
 
 [[package]]
-name = "aho-corasick"
-version = "0.7.18"
+name = "ahash"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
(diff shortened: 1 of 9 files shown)


SpaceManiac:
Any particular motive?

I'm not excited to accept this as-is because even in a skim I see various line wrappings/unwrappings that rustfmt does that straight up make the code less readable. I'd have to spend time to sit down with it to figure out how to make it not touch those.

jupyterkat:
It autoformats your code, for spending less time pressing tabs and enters. And it makes the code more readable most of the time. Only in some cases like the fast table fmt that it screws up.

willox:
This always annoyed me when working on SpacemanDMM stuff since, like many other Rust coders, I just throw shit at the wall and format afterwards

Can't do that here since it changes unrelated code :)

jupyterkat:
those one offs can be easily fixed with an attribute y'know

SpaceManiac:
Having seriously attempted it, this diff is too big to review. I'm happy to accept just the dependency upgrades if you can tolerate separating them out from this PR, but I will rustfmt the project myself if and when I am convinced about it.

SpaceManiac (approved):

PR #352, opened 2023-04-06
Update dreammaker crate readme to use proper subdomain

out-of-phaze:
Lummox says you shouldn't use the `secure` subdomain for the BYOND forums.


SpaceManiac:
Makes sense, but do you have a reference?

ike709:
image

SpaceManiac:
So what's the rule, is it just the forums or should I be replacing every reference to `secure.byond.com`? Is there a _public_ reference?

For the record, all these links were added before `www.byond.com` was even available over HTTPS.

PR #353, opened 2023-04-14
Add call_ext()() support for BYOND 515

willox:
This should be batched with the other 515 changes that are necessary since this can't really ship until DM_VERSION bumps up to 515.


SpaceManiac:
> should be batched

Not bothering to branch because I don't foresee any maintenance releases needed between now and 515 support

PR #356, opened 2023-05-06
Should not sleep Fixes

### LINKED ISSUE #355: Missed sleep with SHOULD_NOT_SLEEP
This errors in dreamchecker as expected
```dm
/datum/proc/is_valid(value)
    SHOULD_NOT_SLEEP(TRUE)

/datum/choiced/is_valid(value)
    get_choices()

/datum/choiced/proc/get_choices()
    init_possible_values()

/datum/choiced/proc/init_possible_values()

/datum/choiced/ai_core_display/init_possible_values()
    sleep(1)
```
This doesn't
```dm
/datum/proc/is_valid(value)
    SHOULD_NOT_SLEEP(TRUE)

/datum/choiced/is_valid(value)
    get_choices()

/datum/choiced/proc/get_choices()
    init_possible_values()

/datum/choiced/proc/init_possible_values()

/datum/choiced/ai_core_display/init_possible_values()
    stoplag()

/proc/stoplag()
    sleep(1)
```

Cyberboss:
Fixes #355 

Also splits SHOULD_NOT_SLEEP and SHOULD_BE_PURE tests and adds a minor improvement to the test helpers in that it shows more info where it wouldn't when `panic`ing


crates/dreamchecker/src/lib.rs
```diff
@@ -562,7 +562,6 @@ pub struct AnalyzeObjectTree<'o> {
     impure_procs: ViolatingProcs<'o>,
     waitfor_procs: HashSet<ProcRef<'o>>,
 
-    sleeping_overrides: ViolatingOverrides<'o>,
```
Cyberboss, line 565: The way this was implemented wasn't effective and the cause of the misses because it didn't catch overrides that called that called procs that slept.

crates/dreamchecker/src/lib.rs
```diff
@@ -651,16 +649,22 @@ impl<'o> AnalyzeObjectTree<'o> {
...
-                    let mut callstack = CallStack::default();
-                    callstack.add_step(*proccalled, *location, *new_context);
-                    to_visit.push_back((*proccalled, callstack, *new_context));
+            let mut to_visit = VecDeque::<(ProcRef<'o>, CallStack, bool, ProcRef<'o>)>::new();
+
+            visited.insert(*procref);
+
+            if let Some(calledvec) = self.call_tree.get(procref) {
+                for (proccalled, location, new_context) in calledvec.iter() {
+                    let mut newstack = CallStack::default();
```
Cyberboss, line 660: This loop is now slow as this on /tg/, we need to cache these callstacks on the procs we check.

crates/dreammaker/src/objtree.rs
```diff
@@ -266,6 +266,17 @@ impl<'a> TypeRef<'a> {
         }
     }
 
+    /// Recursively visit this and all child **types**.
+    pub fn recurse_types<F: FnMut(TypeRef<'a>)>(&self, f: &mut F) {
+        self.recurse(f);
+        for parent_typed_idx in &self.tree.redirected_parent_types {
```
SpaceManiac, line 272: this is going to visit things multiple times

Cyberboss: Is it? `redirected_parent_types` should only be populated with things that have different `parent_type`s from their default path. So if `recurse` only hits direct child paths I don't see how it'd hit things multiple times.

Cyberboss:
As it stands, this is still too slow and uses way too much memory. More improvements incoming.

Cyberboss:
From Discord:

Okay @SpaceManiac#6618 This is latest. The 7228 is the number of procs infected with SHOULD_NOT_SLEEP to begin with. Each print-out represents 10s of work.
```
7228 procs to anaylze call chains for
serialized objtree in 1.718s
11/7228 procs anaylzed
165/7228 procs anaylzed
449/7228 procs anaylzed
818/7228 procs anaylzed
1273/7228 procs anaylzed
1775/7228 procs anaylzed
2250/7228 procs anaylzed
2792/7228 procs anaylzed
3313/7228 procs anaylzed
3840/7228 procs anaylzed
4339/7228 procs anaylzed
4800/7228 procs anaylzed
5349/7228 procs anaylzed
6045/7228 procs anaylzed
6610/7228 procs anaylzed
7109/7228 procs anaylzed
dreamchecker 181.200s - total 192.546s
```

At this point the only optimization I can think of is to NEVER revisit procs (currently procs that _don't_ have a sleep somewhere in their calls can get revisited) It would result in a slightly incomplete error set but it might be worth it.

Cyberboss:
Okay @SpaceManiac This is latest. The 7228 is the number of procs infected with SHOULD_NOT_SLEEP to begin with. Each print-out represents 10s of work.
```
7228 procs to anaylze call chains for
serialized objtree in 1.718s
11/7228 procs anaylzed
165/7228 procs anaylzed
449/7228 procs anaylzed
818/7228 procs anaylzed
1273/7228 procs anaylzed
1775/7228 procs anaylzed
2250/7228 procs anaylzed
2792/7228 procs anaylzed
3313/7228 procs anaylzed
3840/7228 procs anaylzed
4339/7228 procs anaylzed
4800/7228 procs anaylzed
5349/7228 procs anaylzed
6045/7228 procs anaylzed
6610/7228 procs anaylzed
7109/7228 procs anaylzed
dreamchecker 181.200s - total 192.546s
```

At this point the only optimization I can think of is to NEVER revisit procs (currently procs that _don't_ have a sleep somewhere in their calls can get revisited) It would result in a slightly incomplete error set but it might be worth it.

Cyberboss:
I'm going to go ahead and do that

Cyberboss:
Back under 5s. It no longer prints out transitive errors but the time save is worth it.

Cyberboss:
There's a bug where the checking in `/New()` is too aggressive (We KNOW the type at that point and shouldn't be checking overrides).

Cyberboss:
`dreamchecker 2.882s - total 18.013s`

Cyberboss:
If anyone wants to take over this PR. It's basically done, except you need to get the commented out test from the above commit working

crates/dreamchecker/tests/sleep_tests.rs
```diff
@@ -258,6 +258,7 @@ fn sleep8() {
 }
 
 // Testing avoidance of false positive from overrides of different type chains
+/* This test _should_ be enabled but it currently fails because I'm not sure if there's a way to know if we're calling from src or another var
```
LemonInTheDark, line 261: is this the difference between call() and thing.call()?

Cyberboss: yes

Cyberboss:
Superseded by #472

PR #357, opened 2023-05-06
Optimize as much as possible in the release build

Cyberboss:
At ~~the cost of extravagant compile times~~ basically zero cost.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,4 @@ opt-level = 2
 
 [profile.release]
 lto = true
+codegen-units = 1


Cyberboss:
Maybe a third profile only used for releases?

SpaceManiac:
How much is the runtime speedup? How much is the compile slowdown?

Cyberboss:
I'm running a 4 core i5 3670K @3.78GHz

`PS S:\workspace\SpacemanDMM> cargo clean; cargo build --release`

Clean compile times without: `Finished release [optimized] target(s) in 5m 55s`
Clean compile times with: `Finished release [optimized] target(s) in 5m 26s`

My caches must have been warming up with the first test, but I think with those results it's safe to say the difference is negligible.

Using Dreamchecker as the example (ran 3 times in a row each to warm the cache, showing third run results):

`PS S:\workspace\tgstation> Measure-Command { ..\SpacemanDMM\target\release\dreamchecker.exe | Out-Default }`

Without: `TotalSeconds      : 10.0468365`
With: `TotalSeconds      : 9.5563742`

Seems worth it.

PR #366, opened 2023-10-02
Implements base functionality for __TYPE__ and __PROC__

LemonInTheDark:
We basically just capture them early, and then treat them as unique symbols. We don't replace inline or anything because the required context isn't quite there.

I'm not really sure I have ctrl click functionality working. To be honest with you I'm not really sure how to hook into that system, pointers would be appreciated. I thought it was based off the use of use_symbol in find refs but that doesn't seem to work.

### DIFF
--- a/crates/dm-langserver/src/find_references.rs
+++ b/crates/dm-langserver/src/find_references.rs
@@ -544,6 +544,20 @@ impl<'o> WalkProc<'o> {
                     StaticType::None
                 }
             },
+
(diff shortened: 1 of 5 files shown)


SpaceManiac:
> I'm not really sure I have ctrl click functionality working

Fine to defer this to later. `use_symbol` is for Find All References; Ctrl+Click is handled in `on GotoDefinition`.

PR #367, opened 2023-10-03
Implements the scope operator

LemonInTheDark:
Adds support for :: to the parser, in all its forms (global proc/var, off type)
Also implements behavior for it. We'll properly read the type of what we attach to, and double check that everything matches. Works for the proc refs too.

I've added unit tests for all this to double check my work. I DIDN'T mirror the static var detection byond does. problem for another day.

Please let me know if the way I'm going about this is silly/mislead, I'll declutter as directed


LemonInTheDark:
I tell a lie, it works fine

PR #369, opened 2023-10-04
Implements /proc/final, piped into should not override

LemonInTheDark:
As a part of this we need to make a struct when building proc defs, to keep track of their flags AND their kind at the same time.

We then pass that down into a new flags var on the proc decl struct, and we're golden

Only thing of note is I removed the is_private and is_protected vars from proc declerations because they were totally unused. Spooky added em a long time ago and I think it was for not much reason


LemonInTheDark:
Fixing proc return types, please hold

LemonInTheDark:
Moved the as syntax stuff to #370, this is all clear

LemonInTheDark:
sorry accidentally killed your commit, it's still built in tho. had trouble with two different locals and forgot which had most recent changes.

PR #371, opened 2023-10-05
Implement %%, %%=, and operator""

LemonInTheDark:
The float stuff was easy. Operator was too, just took a bit to realize how it worked

Nothin much else to say, outside that I needed to roll my own operation for this one, since float remainder is kinda hard to find (and I couldn't after cursory reading)


crates/dreammaker/src/parser.rs
```diff
@@ -944,6 +950,8 @@ impl<'ctx, 'an, 'inp> Parser<'ctx, 'an, 'inp> {
             } else {
                 last_part.push_str("[]");
             }
+        } else if self.exact(Punct(EmptyQuote))?.is_some() {
```
SpaceManiac, line 953: A less shocking change might be to look for an empty `Token::String` here.

crates/dreammaker/src/lexer.rs
```diff
@@ -71,9 +71,12 @@ table! {
     "!",   Not;
     "!=",  NotEq;
     "\"",  DoubleQuote;
+    "\"\"",  EmptyQuote;
```
SpaceManiac, line 74: Now that this has its own token, the parser no longer recognizes it.

SpaceManiac (changes requested):

crates/dreammaker/src/lexer.rs
LemonInTheDark: Oh, I see. I was unsure how the token system actually worked. 
I'll see about using Token::string instead, cause that's bad you're right

crates/dreammaker/src/parser.rs
LemonInTheDark: handled, I think?

PR #372, opened 2023-10-06
Implements parsing for #pragma, handles the multiple case niavely

LemonInTheDark:
I'm using a second hashmap to track multiple'd files. Not sure that this is the best way of going about it, but it's a living

### DIFF
--- a/crates/dreammaker/src/preprocessor.rs
+++ b/crates/dreammaker/src/preprocessor.rs
@@ -104,6 +104,7 @@ impl DefineHistory {
             env_file: self.env_file.clone(),
             include_stack: Default::default(),
             include_locations: Default::default(),
+            multiple_locations: Default::default(),
             history: Default::default(),  // TODO: support branching a second time
             defines,
             maps: Default::default(),
@@ -129,6 +130,7 @@ impl DefineHistory {
             env_file: self.env_file.clone(),
             include_stack: Default::default(),
             include_locations: Default::default(),
+            multiple_locations: Default::default(),
             history: Default::default(),  // TODO: support branching a second time
             defines: DefineMap::from_history(self, self.last_input_loc),
             maps: Default::default(),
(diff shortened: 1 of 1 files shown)


crates/dreammaker/src/preprocessor.rs
```diff
@@ -952,6 +967,18 @@ impl<'ctx> Preprocessor<'ctx> {
                         expect_token!((text) = Token::String(text));
                         self.context.register_error(DMError::new(self.last_input_loc, format!("#{} {}", ident, text)));
                     }
+                    "pragma" if disabled => {}
+                    "pragma" => {
+                        expect_token!((text) = Token::String(text));
```
SpaceManiac, line 972: You probably mean Ident here, not String

crates/dreammaker/src/preprocessor.rs
```diff
@@ -952,6 +967,18 @@ impl<'ctx> Preprocessor<'ctx> {
...
+                        match text.as_str() {
+                            "multiple" => {
+                                if let Some(file_id) = self.context.get_file(self.env_file.as_path()) {
```
SpaceManiac, line 975: `self.env_file` is the root, not the file currently being processed

SpaceManiac (changes requested):

crates/dreammaker/src/preprocessor.rs
LemonInTheDark: I changed it, but are you sure? #warn uses String() and its input isn't "'d at all, it's just the text after the #warn

SpaceManiac: It's stupid, but there's a hack in lexer.rs for this: `if ident == "warn" || ident == "error" {`

LemonInTheDark: in fucking tears

LemonInTheDark: should I add to that or just keep as is

SpaceManiac: As written is fine. The reason is that #warn and #error want to treat the whole rest of the line as one big string. #pragma is (for now...) followed by ordinary tokens.

LemonInTheDark: alright. is there anything else you want me to do here?

PR #379, opened 2023-11-16
dmm_tools::dmm::Map::adjust_key_length() fix

DreamySkrell:
Currently, there is `Map::adjust_key_length()`, but it sets key length based on the amount of items in the dictionary. Which is fine if keys in order like 1-2-3-4-...-N, but it falls apart if the keys are not in order - which is valid for a dmm map.

If a map has keys like [1, 2, 3000], then, currently, `Map::adjust_key_length()` will set key length to 1, but the map will panic on save cause the 3000 key is bigger than max key for this key length (52).

This PR fixes `adjust_key_length` to work properly in that case, where it sets key length based on the biggest key in the map dictionary.

### DIFF
--- a/crates/dmm-tools/src/dmm.rs
+++ b/crates/dmm-tools/src/dmm.rs
@@ -191,12 +191,15 @@ impl Map {
     }
 
     pub fn adjust_key_length(&mut self) {
-        if self.dictionary.len() > 2704 {
-            self.key_length = 3;
-        } else if self.dictionary.len() > 52 {
-            self.key_length = 2;
-        } else {
-            self.key_length = 1;
+        if let Some(max_key) = self.dictionary.keys().max() {
+            let max_key = max_key.0;
+            if max_key > 2704 {
+                self.key_length = 3;
+            } else if max_key > 52 {
+                self.key_length = 2;
+            } else {
+                self.key_length = 1;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


SpaceManiac:
Perhaps this function should validate that the passed-in length is 1, 2, or 3.

Perhaps `adjust_key_length` should also be fixed!

DreamySkrell:
> Perhaps this function should validate that the passed-in length is 1, 2, or 3.
> 
> Perhaps `adjust_key_length` should also be fixed!

Now thinking about it, fixing `adjust_key_length` would be better and safer, where it would iterate over all dict keys and get the largest key, and set key length based on that.

Even with the validation for `set_key_length` as written currently, the user could still give it 1 when there's too many keys, and it would panic on save.

DreamySkrell:
Updated code and PR as above.

crates/dmm-tools/src/dmm.rs
```diff
@@ -194,12 +194,15 @@ impl Map {
...
-            self.key_length = 3;
-        } else if self.dictionary.len() > 52 {
-            self.key_length = 2;
-        } else {
-            self.key_length = 1;
+        if let Some(max_key) = self.dictionary.keys().max() {
+            let max_key = max_key.0;
+            if max_key > 2704 {
+                self.key_length = 3;
+            } else if max_key > 52 {
```
SpaceManiac, line 201: I think these two conditions should now be >= instead of just >

DreamySkrell: addressed in most recent commit

PR #387, opened 2024-05-15
Adds delay argument to animate

Majkl-J:
515.1590 added the delay argument to animate() and I happen to be using it


ZeWaka:
image
yep

PR #388, opened 2024-05-24
fexists can be called in the preprocessor

ZeWaka:
title, 515 feature https://www.byond.com/forum/post/2857912

currently, 
image

can't actually const eval since it can have diff. behavior at runtime obv

### DIFF
--- a/crates/dreammaker/src/constants.rs
+++ b/crates/dreammaker/src/constants.rs
@@ -815,11 +815,14 @@ impl<'a> ConstantFolder<'a> {
                         None => return Err(self.error("malformed nameof() call, expression appears to have no name")),
                     }
                 }
-                "fexists" => {
+                "fexists" if self.defines.is_some() => {
                     if args.len() != 1 {
                         return Err(self.error(format!("malformed fexists() call, must have 1 argument and instead has {}", args.len())));
                     }
-                    Constant::Call(ConstFn::FileExists, self.arguments(args)?)
+                    match args[0].as_term() {
+                        Some(Term::String(ref path)) => Constant::from(std::path::Path::new(path).exists()),
+                        _ => return Err(self.error("malformed fexists() call, argument given isn't a string.")),
+                    }
                 }
                 // other functions are no-goes
                 _ => return Err(self.error(format!("non-constant function call: {}", ident))),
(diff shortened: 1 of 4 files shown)


crates/dreammaker/src/constants.rs
```diff
@@ -812,6 +815,12 @@ impl<'a> ConstantFolder<'a> {
                         None => return Err(self.error("malformed nameof() call, expression appears to have no name")),
                     }
                 }
+                "fexists" => {
+                    if args.len() != 1 {
+                        return Err(self.error(format!("malformed fexists() call, must have 1 argument and instead has {}", args.len())));
+                    }
+                    Constant::Call(ConstFn::FileExists, self.arguments(args)?)
```
SpaceManiac, line 822: I suggest:
1. Like `"defined"`, requiring `self.defines.is_some()`, so that this function will be properly rejected if used in a type var context.
2. Like `"defined"`, actually check whether the file exists and evaluate to 0 or 1 here. Otherwise a `Constant::Call(..)` is always considered truthy (see `Constant::to_bool`) and `#if fexists("anything")` will always pass.

crates/dreammaker/tests/constants_tests.rs
```diff
@@ -110,3 +110,11 @@ fn rgb_hcy() {
...
     );
 }
+
+#[test]
+fn fexists_noarg() {
+    assert_eq!(
+        eval("fexists()").unwrap_err().description(),
+        "malformed fexists() call, must have 1 argument and instead has 0",
+    );
+}
```
SpaceManiac, line 120: True/false tests would be good here too.

SpaceManiac (changes requested):

crates/dreammaker/tests/constants_tests.rs
ZeWaka: Couldn't get tests working, due to either not evaluating the preprocessor or the ability to create files within the tests.

crates/dreammaker/src/constants.rs
ZeWaka: should be addressed

crates/dreammaker/tests/constants_tests.rs
ZeWaka: guh, those didn't work when i tried, oh well ty

PR #398, opened 2024-07-07
Changes severity of the `redefined proc` to Warning

DGamerL:
These can be a lot of trouble and should be a lot more visible to fix
This will almost certainly create a lot of warnings on many codebases, not sure if I can/should do something for that?


SpaceManiac:
146 detections in /tg/, many of which are on purpose.P I think it will have to stay a hint by default.

To make it a warning just for your codebase, add this to SpacemanDMM.toml adjacent to the .dme file:
```toml
[diagnostics]
redefined_proc = "warning"
```

DGamerL:
Did not know that was a thing, apologies!
Most of the errors seem to be because of global vars being protected, which I wouldn't quickly know how to fix anyways!

PR #401, opened 2024-07-15
fix preprocessor relative filepath evaluation for fexists

ZeWaka:
The way I implemented const-eval'd fexists is slightly incorrect.
It's evaulating the existence of the path relative to the invocation of the library or something, instead of relative to the ~~dme~~ current file.
This is causing errors in StrongDMM, because it's checking paths relative to the application exe file.

I tested by building StrongDMM and using this fixed version, and it worked correctly.

### DIFF
--- a/crates/dreammaker/src/constants.rs
+++ b/crates/dreammaker/src/constants.rs
@@ -2,6 +2,7 @@
 use std::fmt;
 use std::ops;
 use std::path::Path;
+use std::path::PathBuf;
 
 use get_size::GetSize;
 use get_size_derive::GetSize;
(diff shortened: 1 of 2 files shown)


crates/dreammaker/src/constants.rs
```diff
@@ -444,8 +446,9 @@ impl Expression {
 }
 
 /// Evaluate an expression in the preprocessor, with `defined()` available.
-pub fn preprocessor_evaluate(location: Location, expr: Expression, defines: &DefineMap) -> Result<Constant, DMError> {
+pub fn preprocessor_evaluate(location: Location, expr: Expression, defines: &DefineMap, env_file: Option<&PathBuf>) -> Result<Constant, DMError> {
```
ZeWaka, line 449: This is technically an API break because it's a public function I think? I made it an Option but i think outside calls would still technically scream about the new argument.

SpaceManiac: No problem

PR #408, opened 2024-08-31
Coalesce duplicate tile definitions on save.

warriorstar-orion:
This change modifies maps just before saving to find instances of multiple keys sharing the same prefab data, and removing all but one from the dictionary, assigning that one to any coordinate on the grid which needs that prefab data.

No test case but I ran the implementation through https://github.com/warriorstar-orion/avulto with small and large maps and sanity checked its behavior. It doesn't seem to have much of a performance impact on save, either.

### DIFF
--- a/crates/dmm-tools/src/dmm.rs
+++ b/crates/dmm-tools/src/dmm.rs
@@ -1,4 +1,5 @@
-use std::collections::BTreeMap;
+use std::borrow::BorrowMut;
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt;
 use std::fs::File;
 use std::io;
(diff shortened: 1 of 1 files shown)


LemonInTheDark:
what causes this behavior in the first place?

warriorstar-orion:
> what causes this behavior in the first place?

This is basically a guard against programmer error; it's possible through normal use of the library to generate new keys and assign them prefab data that already exists elsewhere in the mapping, and it's not pragmatic to try and track that yourself. For example, if you wrote a program that adds an /atom/foo to every tile on a map if it doesn't already have one, and you have one tile with the contents [/turf, /area] and one with the contents [/turf, /area, /atom/foo], the first tile will end up with the same contents as the second, but you don't want to perform that check for every operation you perform on a tile on the off chance that you may have created two duplicate mappings. It's easier and more consistent instead for the library to handle that for all cases.

LemonInTheDark:
Oh I see, misunderstood, yeah that tracks

PR #412, opened 2024-10-25
Upgrade packages to Rust 2021

ZeWaka:
More modern edition of the compiler

### DIFF
--- a/crates/dm-langserver/src/macros.rs
+++ b/crates/dm-langserver/src/macros.rs
@@ -142,7 +142,7 @@ macro_rules! if_annotation {
 }
 
 macro_rules! match_annotation {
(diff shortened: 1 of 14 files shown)


crates/dm-langserver/src/macros.rs
```diff
@@ -142,7 +142,7 @@ macro_rules! if_annotation {
 }
 
 macro_rules! match_annotation {
-    ($a:expr; $($($p:pat)|* => $b:block,)*) => {
+    ($a:expr; $($($p:pat_param)|* => $b:block,)*) => {
```
ZeWaka, line 145: https://doc.rust-lang.org/edition-guide/rust-2021/or-patterns-macro-rules.html#summary

PR #414, opened 2024-11-02
Change the hasher used for HashMap/Sets

ZeWaka:
I chose to use [foldhash](https://docs.rs/foldhash/0.1.3/foldhash/index.html) since we're not needing cryptographically secure hashing, nor are we concerned with people reverse-engineering application state.

Basically just changing the RandomState or the HashMap/Set import everywhere. The *Ext types provide the `::new()` interface.

Tested dm-langserver, all works well:
image


SpaceManiac (approved):
Also seeing a clear 7-9% speedup on a couple of machines. Binary size is not increased.

PR #418, opened 2025-01-06
Two micro-opts for lexer

tigercat2000:
Measured about ~150ms faster peek() with the first change.
Measured within-noise for the second change- but it's nicer code so it's still a good change, I think.

### DIFF
--- a/crates/dreammaker/src/lexer.rs
+++ b/crates/dreammaker/src/lexer.rs
@@ -935,20 +935,22 @@ impl<'ctx> Lexer<'ctx> {
         }
     }
 
-    fn read_ident(&mut self, first: u8) -> String {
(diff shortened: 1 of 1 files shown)


SpaceManiac:
150ms must be a typo; did you mean ns? The speedup is within noise parsing an entire codebase on a release build. It seems to me that on `-C opt-level=2` or higher, where PartialEq gets inlined, the assembly has a len== check anyways.

Improving the put_back code is good, but you could go even further by moving the `let ws = ...` inside `read_ident` and return `bool` instead of `Option<u8>`.

tigercat2000:
Honestly I think I profiled in debug mode by mistake lol

SpaceManiac:
I appreciate the `read_ident` suggestion all the same

PR #420, opened 2025-02-28
Add 516 compatibility stubs + astype

hry-gh:
we still need to do `for (k, v in list()/alist())` and `pragma syntax c` but it'll do for now

### DIFF
--- a/crates/dreammaker/src/builtins.rs
+++ b/crates/dreammaker/src/builtins.rs
@@ -13,8 +13,8 @@ const DM_BUILD: i32 = 1619;
 
 /// Register BYOND builtin macros to the given define map.
 pub fn default_defines(defines: &mut DefineMap) {
-    use super::lexer::*;
     use super::lexer::Token::*;
+    use super::lexer::*;
     let location = Location::builtins();
 
     // #define EXCEPTION(value) new /exception(value)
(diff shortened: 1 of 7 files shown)


LemonInTheDark:
Missing a few other things, didn't see this before you're so epic and cool

crates/dreammaker/src/builtins.rs
```diff
@@ -629,6 +636,22 @@ pub fn register_builtins(tree: &mut ObjectTreeBuilder) {
         list/proc/Swap(Index1, Index2);
         list/var/len;
 
+        // 516
+        alist;
+        alist/var/const/type;
+        alist/var/const/parent_type;
+        alist/var/tag;
```
LemonInTheDark, line 643: this isn't a datum child, is it?

hry-gh: LXdMDbpbOaeM7y6D@2x

wouldn't that be nice 🙁

LemonInTheDark: hghhhhhhhhhhhhhhhhh

hry-gh: the answer is not really, it just exposes a tag var (for some really awful usecases ig) and a const parent/type var, like lists

SpaceManiac:
It seems to me that `load_ext` and `astype` both syntactically resembles standard proc calls (unlike `call_ext` which has TWO pairs of parentheses after it instead of one) so I think they don't need their own special `Term` entries.

Otherwise looks good to me.

hry-gh:
> It seems to me that `load_ext` and `astype` both syntactically resemble standard proc calls (unlike `call_ext` which has TWO pairs of parentheses after it instead of one) so I think they don't need their own special `Term` entries.
> 
> Otherwise looks good to me.

done!

PR #425, opened 2025-04-04
new() works for alist, new primitives. <=> and stubs bound_pixloc()

hry-gh:


SpaceManiac:
`<=>` will need to be added to the overloadable operator set in `try_read_operator_name`

hry-gh:
> `<=>` will need to be added to the overloadable operator set in `try_read_operator_name`

done!

PR #431, opened 2025-06-04
Basic support for `for (var/k, v in X)` syntax

ShiftyRail:
## What this does 

This add a new arm to the parser for the `for (var/k, v in X)` dict-like syntax that was added in 516.

## How this works

I have hijacked the arm that checks for this type of expressions : 

`for (init, test, inc)` or `for (init; test; inc)`

I have seperated it between `init; test; inc` (left untouched) and `init, test, inc`.
If `inc` is not None, I then check if `test` in a `BinaryOp::In` (something like `for(init, [x in y])`. If not I just assume it's a `for (init, test)` which is valid DM syntax.

I then check if `init` is a `var/k` type of statement, using copypaste. DM will not compile `for (k, v)` nor will it compile `for (var/k, var/v)`. 
I check if v is an ident.

If we're all good, I create a new type of structure `ForKeyValueStatement` and the lib.rs / findreferences.rs knows what to do with it.

## Cases tested : 

<details>
<summary>See unit tests in the Files Changed</summary>

```dm
for (var/k, v in L)
     world.log << k
     world.log << v
```
✅

```dm
for (k, v in L)
     world.log << k
     world.log << v
```
❌ "for (var/key, value) requires a 'var' keyword"

```dm
for (var/k = 1, v in L)
     world.log << k
     world.log << v
```
❌  "cannot assigned a value in var/key for(var/key, value) statement"

```dm
for (var/0, v in L)
     world.log << k
     world.log << v
```
❌  "path has no effect" (was already covered)

```dm
for (var/k, 0 in L)
     world.log << k
     world.log << v
```
❌ "value must be a variable in for (var/key, value) statement"
</details>

## Regression checks 

Both `for (i = 0, i < 10, i++)`, `for (i = 0, i < 10)` still pass.

### DIFF
--- a/crates/dm-langserver/src/find_references.rs
+++ b/crates/dm-langserver/src/find_references.rs
@@ -325,7 +325,8 @@ impl<'o> WalkProc<'o> {
             Statement::Goto(_) => {},
             Statement::Crash(_) => {},
             Statement::Label { name: _, block } => self.visit_block(block),
-            Statement::Del(expr) => { self.visit_expression(location, expr, None); },
+            Statement::Del(expr) => { self.visit_expression(location, expr, None); }, // UUuuh
+            Statement::ForKeyValue(_) => { println!("correctly found a kv pair but no idea what do lol!")},
         }
     }
 
--- a/crates/dm-langserver/src/find_references.rs
+++ b/crates/dm-langserver/src/find_references.rs
@@ -333,8 +333,15 @@ impl<'o> WalkProc<'o> {
                 }
                 if let Some(var_type) = var_type {
                     self.visit_var(location, var_type, key, None);
-                    self.visit_var(location, var_type, value, None);
                 }
+                // the "v" in a DM for (var/k, v) statement is essentially typeless.
+                // There is currently no way to change that.
+                let var_type_value = VarType {
+                    flags: VarTypeFlags::from_bits_truncate(0),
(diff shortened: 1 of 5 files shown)


ZeWaka (commented):
Also, don't we have tests? Why put your test cases in the description if you're not going to put in as tests?

ShiftyRail:
I agree with the var_type thing - however DM will not let you give a var type to v in the `for(var/k, v`. 

Thinking about it I should create a `var_type` corresponding to an empty typepath and use that instead but I am not really sure how I can actually do that.

I didnt notice the parser had unit tests. I am more than happy to include the `for(var/k, v` cases. I included them because I wanted to show maintainers the cases that I had tested in the event that they had questions about it :)

ShiftyRail:
I've added a basic unit test that shows a ```for(var/k, v in A)``` is parsed. I can add more specific error messages checking with some help -- I have no idea how to make rust check exactly that an error message got printed.

I've also added a block :
```rust
                // the "v" in a DM for (var/k, v) statement is essentially typeless.
                // There is currently no way to change that.
                let var_type_value = VarType {
                    flags: VarTypeFlags::from_bits_truncate(0),
                    type_path: Box::new([]),
                    input_type: InputType::from_bits_truncate(0),
                };
```

DM lets you type `k` but values are always essentially typeless and there's no way to change it as the moment.

PR #433, opened 2025-06-17
Add support for `call_ext(LoadedFunc)` syntax

Absolucy:
this adds support for 516's `call_ext(LoadedFunc)` syntax (allowing either one or two args to `call_ext`)

the `call_ext_missing_arg` test is removed as a result, as `call_ext` with a single arg is now valid syntax.

### DIFF
--- a/crates/dreammaker/src/ast.rs
+++ b/crates/dreammaker/src/ast.rs
@@ -1071,8 +1071,8 @@ pub enum Term {
     DynamicCall(Box<[Expression]>, Box<[Expression]>),
     /// A use of the `call_ext()()` primitive.
     ExternalCall {
-        library_name: Box<Expression>,
-        function_name: Option<Box<Expression>>,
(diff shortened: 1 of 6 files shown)


crates/dreammaker/src/ast.rs
```diff
@@ -1072,7 +1072,7 @@ pub enum Term {
     /// A use of the `call_ext()()` primitive.
     ExternalCall {
         library_name: Box<Expression>,
-        function_name: Box<Expression>,
+        function_name: Option<Box<Expression>>,
```
SpaceManiac, line 1075: Maybe these should be renamed. It's not really a "library name" in the single-argument form.

Absolucy: what would be a good name for it?
or would it perhaps be a good idea to just make a separate `ExternalCall` and `LoadedExternalCall` variants?

PR #437, opened 2025-07-13
Booleans are now evaluated as their own Term, instead of being rendered as an Integer/Float

Metekillot:
This PR adds booleans as an AST element, replacing the implementation where they're represented as literal 1 or 0 either as integers or floats.


SpaceManiac:
Reasoning? The way it is now matches how BYOND does it.

Metekillot:
> Reasoning? The way it is now matches how BYOND does it.

It made the syntax tree easier for me to understand and so I thought it would be helpful!

SpaceManiac:
Ah, I see. Unfortunately in this case I see it wise to keep parity with BYOND and not have a separate type for something it does not. This is the same reason we use only have Floats and do not keep a separate Int type.

PR #438, opened 2025-07-14
Add an Option<Range<Location>> body_range field to ProcValue, for concise and lighter-weight reference

Metekillot:
I wasn't able to find a concise way to reference the range and/or contents of proc bodies without also needing to reference the Annotation tree. I haven't implemented any use case for the added information outside of some experiments I'm doing with the deeper analysis companion; I think it could be useful for determining return values and types without the need for SpacemanDMM directives, at a later point.

It's an Option to account for built-ins not having any actual presence in the syntax tree.

### DIFF
--- a/crates/dreammaker/src/objtree.rs
+++ b/crates/dreammaker/src/objtree.rs
@@ -2,6 +2,8 @@
 
 use std::collections::BTreeMap;
 use std::fmt;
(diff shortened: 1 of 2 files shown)


SpaceManiac:
Seems useful enough.

PR #439, opened 2025-07-19
Add the missing sign(A) builtin proc

FunnyMan3595:
It's not tagged with a version in the DM reference, so I assume it's always been there, but got missed when this list was created.


FunnyMan3595:
Ack, yeah, and I see the 516 label hiding way up in the corner of the reference page.

PR #447, opened 2025-12-09
fix parsing of alist for-k,v key type filtering

ZeWaka:
fix parsing of
```dm
var/alist/foo = list()
for (var/index as num, C in foo)
	world << index
```
currently `got ',', expected one of: '|', 'in', ')'`

also fixes two clippy warnings

ref: bottom of https://www.byond.com/docs/ref/#/proc/for/list


crates/dreammaker/src/config.rs
```diff
@@ -86,10 +86,12 @@ pub enum WarningLevel {
 
 /// Available debug engines.
 #[derive(Debug, Deserialize, Clone, Copy, PartialEq)]
+#[derive(Default)]
```
ZeWaka:
```suggestion
#[derive(Debug, Default, Deserialize, Clone, Copy, PartialEq)]
```
actually this should just be on one line lol

PR #449, opened 2026-01-08
replace linear algebra types

salixsalicaceae:
Refactored `Coord2` and `Coord3` to instead use `glam` types, so as to reduce potential overhead from maintaining i32 vector types that have already been implemented. Struct identifiers were retained, so as to enable implementation of operations that are specific to the scope of BYOND.


SpaceManiac:
I don't think I see a meaningful benefit to this.

PR #461, opened 2026-05-04
Rust 2024 apparently?

Metekillot:
At this time I'm honestly kind of in an uneasy daze because this was much more straightforward than I anticipated. I assume I'm missing some important build step and I'm actually building in 2021 still.

In any case, I just changed all the Cargo.toml files to use the 2024 edition then made the changes the compiler told me to until it compiled. All the tests are passing as well.

### DIFF
--- a/CONFIGURING.md
+++ b/CONFIGURING.md
@@ -31,6 +31,7 @@ Raised by DreamChecker:
 * `override_missing_keyword_arg` - Raised when proc overrides are missing keyword arguments
 * `must_not_override` - `SpacemanDMM_should_not_override` directive
 * `must_call_parent` - `SpacemanDMM_should_call_parent` directive
+* `must_not_call_parent` - `SpacemanDMM_should_not_call_parent` directive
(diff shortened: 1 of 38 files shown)


SpaceManiac:
Directionally correct, but:
- Motivation?
- Odd whitespace
- Still a compile error somewhere

Metekillot:
> Directionally correct, but:
> - Motivation?

I'm making a heavily modified fork as a submodule for interoperability with an LSP CLI tool I'm making, and I'm using Rust 2024 on it; I wanted to minimize friction for anything I'd PR upstream or pull from it when it's aligned with base SDMM functionality.

Rust 2024 seems like it does some things differently with syntax and unsafe blocks and, I think other things too for some standard library stuff; I don't know enough about Rust to speak authoritatively.

> - Odd whitespace

I can fix that.

> - Still a compile error somewhere

I will look into this; I'm compiling on Linux but I believe I was running into some issues with MSVC over the gnu compiler for windows cross compiling

PR #472, opened 2026-07-13
Fixes SHOULD_NOT_SLEEP() missing situations where sleep is in fact called

### LINKED ISSUE #355: Missed sleep with SHOULD_NOT_SLEEP

CabinetOnFire:
Based on #356 and finished with some additional tweaks that I noticed while fixing issues I had on /tg/ locally

Fixes https://github.com/SpaceManiac/SpacemanDMM/issues/355

Some additional changes on top of that PRs original changes:
- since /obj, /mob and /turf are not path-descendant from atom they could be missed in some specific cases. Added a new test for that (sleep10) and fixed it via redirected_parent_types. This was also the issue I was running into on /tg/
- Fixes false positives that triggered on Cyberboss version. We now bound the override lookup to the static receiver type.

The rest is similar, the override tracking system is gone and we now walk the call tree instead. Kept the test case from Cyberboss too (test9)

Also kept the split up test files and improved test helpers.

### DIFF
--- a/crates/dreamchecker/README.md
+++ b/crates/dreamchecker/README.md
@@ -62,6 +62,7 @@ be enabled:
 	#define UNLINT(X) SpacemanDMM_unlint(X)
 	#define SHOULD_NOT_OVERRIDE(X) set SpacemanDMM_should_not_override = X
 	#define SHOULD_NOT_SLEEP(X) set SpacemanDMM_should_not_sleep = X
+	#define ALLOWED_TO_SLEEP(X) set SpacemanDMM_allowed_to_sleep = X
 	#define SHOULD_BE_PURE(X) set SpacemanDMM_should_be_pure = X
 	#define PRIVATE_PROC(X) set SpacemanDMM_private_proc = X
 	#define PROTECTED_PROC(X) set SpacemanDMM_protected_proc = X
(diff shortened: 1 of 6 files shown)


SpaceManiac:
I pushed some minor fixes. But more importantly, the detection seems to be nondeterministic, fluctuating between 113 and 147 diagnostics between repeat runs on current /tg/. This seems bad.

CabinetOnFire:
@SpaceManiac 

Sorry missed your comment.

this was happening because the shared visited set intentionally bounds the amount of traversal we do, but the hashmap is randomized so it was picking different roots each time; resulting in different things claiming the callees and thus ending up with different numbers.

SpaceManiac:
I suppose it's enough for now. Someday I'd like if the search algorithm wasn't so sensitive to this ordering in the first place.

PR #474, opened 2026-07-26
Implement folding range request

ZeWaka:
uses same #region keyword as ts/c#

image

also some newer version clippy warnings

### DIFF
--- a/crates/dm-langserver/src/main.rs
+++ b/crates/dm-langserver/src/main.rs
@@ -1358,6 +1358,7 @@ impl Engine {
         Completion;
         SignatureHelpRequest;
         DocumentSymbolRequest;
+        FoldingRangeRequest;
         DocumentColor;
         ColorPresentationRequest;
         DocumentLinkRequest;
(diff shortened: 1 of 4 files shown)


crates/dm-langserver/src/main.rs
```diff
@@ -2546,6 +2579,23 @@ fn path_to_url(path: PathBuf) -> Result<Url, jsonrpc::Error> {
     Url::from_file_path(path).map_err(|_| invalid_request(format!("bad file path: {formatted}",)))
 }
 
+fn is_region_marker(line: &str, marker: &str) -> bool {
+    let Some(comment) = line.strip_prefix("//") else {
+        return false;
+    };
+
+    // handle doc comments too
+    let rest = comment.strip_prefix('/').unwrap_or(comment);
```
SpaceManiac: I would think that writing `#region` in doc comments currently pollutes the output

ZeWaka: that's fair

ZeWaka: removed doc comment support, and now both
`//#region foo`
and
`// #region foo`
work

SpaceManiac (approved):

PR #479, opened 2026-09-12
Add `should_not_sleep` check version options

ZeWaka:
Adds a new config for `SpacemanDMM_should_not_sleep`:
* v1: pre-#472 behavior
* v2: post-#472, tracks dynamic overrides
* v3: new value provenance to avoid false positives from unknown typed receivers

Basically, only things like `src.proc()`, global/self calls, `..()`, and terms proven to refer to `src`.

v3 ignores the case (which is valid! just very noisy in ss13) of
```dm
/proc/perform(atom/A)
    set SpacemanDMM_should_not_sleep = TRUE
    var/atom/B = A
    B.overridden_proc()

/atom/proc/overridden_proc()
    return

/atom/foo/overridden_proc()
    sleep(1)
```

Goon had hundreds on v2, tg had as well - also just from the first pass, not all errors show at once.
With v3, goonstation has 0 and tgstation has 6 seemingly valid ones. Mostly in martial arts code.


crates/dreamchecker/src/lib.rs
```diff
@@ -745,19 +784,115 @@ impl<'o> AnalyzeObjectTree<'o> {
         }
     }
 
+    fn check_proc_call_tree_legacy(&mut self) {
```
ZeWaka: The old code from before https://github.com/SpaceManiac/SpacemanDMM/pull/472

ZeWaka: Version 3 produces zero errors on goon, so questionable if we even want to have this still. Still 6 errors on tg, though.

crates/dreamchecker/src/lib.rs
```diff
@@ -876,13 +1012,16 @@ impl<'o> AnalyzeObjectTree<'o> {
                         } else {
                             each.src
                         };
+                        let call_receiver_stable =
+                            !receiver_provenance || (each.inherit_receiver && receiver_is_stable);
```
ZeWaka, line 1021: keep receiver provenance only when the call inherits the already-stable receiver

crates/dreamchecker/src/lib.rs
```diff
@@ -745,16 +779,106 @@ impl<'o> AnalyzeObjectTree<'o> {
         }
     }
 
+    fn check_proc_call_tree_legacy(&mut self) {
```
ZeWaka: The old code from before #472
Version 3 produces zero errors on goon and only 6 on tg (seemingly valid ones), so questionable if we even want to have this still.

SpaceManiac:
6-8 diagnostics is much more practical to dump on people than 100+, thanks for the middle ground. I've changed the default from 2 to 3.

